PR #12, opened 2019-04-18
Removing Openssl dependency

lgaches:
- [ ] needs `verify` JWT signature to be implemented at least for testing
- [ ] update README


kylebrowning:
@lgaches unfortunately we cannot remove our OpenSSL dependency until BoringSSL actually becomes complete.

We cant pin to a version of swift-nio-ssl because then we lose security fixes. 

Using BoringSSL has the risk of being rejected as a SSWG package because it's in such a constant fluctuation of it's API.

I do appreciate the work on this commit but it won the accepted until BoringSSL is considered complete in swift

PR #13, opened 2019-04-24
Initial feedback.

kylebrowning:


Sources/NIOAPNS/APNSConfiguration.swift
```diff
@@ -12,6 +12,12 @@ import NIOHTTP2
 import NIOSSL
 
 public struct APNSConfiguration {
+    
+    public enum APNSEnvironment {
```
weissi: nit: I think we could now rename this to just `Environment`. Its full name will be `APNSConfiguraion.Environment`

weissi: btw, if you prefer the types syntactically to be elsewhere, you can also do

```swift
extension APNSConfiguration {
    public enum Environment {
        case production
        case sandbox
    }
}
```

so no need to put it in there. But that's totally up to your preferences

PR #16, opened 2019-05-04
Replaced CAPNSOpenSSL with CNIOBoringSSL

v57:
I think it's better to reuse CNIOBoringSSL from NIOSSL 😉

Tested on macOS 10.14.4 and ubuntu server 16.04


v57:
Oh sorry i just saw pull request #12

PR #26, opened 2019-06-05
Fixed bug

sabeatra:
When the connection fails, responsePromise get's deallocated (deinit is called) without being fulfilled.
This causes a fatal error in debug: "fatalError("leaking promise created at \(creation)", file: creation.file, line: creation.line)", file EventLoopFuture.swift.
The case is that when flatMap fails, there's no path in which the promise get's fulfilled.
This solution fulfills the promise, even if nothing is done with it.

Setup to reproduce:

macOS 1.14.5
Xcode 10.2.1
Swift 5.0.1

Implemented a test website to send push notifications to my phone. There's some other issue that causes an invalidAuthKey error about 50% of the time, which is probably my fault, but it's not the subject of this pull request.

When the error happens, most of the time (sometimes immediately, it's very easy to reproduce) the error described above happens.

I believe there should be a better solution that involves a review of the logic, but as this is my first Digg into the code, I don't feel ready for that, so I hope that you guys review it and tell me if this is a proper solution or there's a cleaner one.

Whatever the case, this fixes the crash (in debug), so I'm going to use it in my own work with my fork.

Please let me know if I can improve this PR as I'm not often participate in open source projects so the etiquette could be off.
Thanks.
Eduardo

### DIFF
--- a/Sources/NIOAPNS/APNSConnection.swift
+++ b/Sources/NIOAPNS/APNSConnection.swift
@@ -80,10 +80,12 @@ public final class APNSConnection {
             request: buffer,
             responsePromise: responsePromise
         )
+
         return streamPromise.futureResult.flatMap { stream in
-            stream.writeAndFlush(context)
-        }.flatMap {
-            responsePromise.futureResult
+            responsePromise.futureResult.whenComplete { _ in }
+            return stream.writeAndFlush(context)
+            }.flatMap {
+                responsePromise.futureResult
         }
     }
 


kylebrowning:
@sabeatra Thanks for the PR. Sorry for any trouble this may have caused you.

I have reviewed, and merged it. Tested on my local machine as well.

Let me know if you figure out your issue with the Invalid Auth key.

sabeatra:
That was fast!

sabeatra:
If you have the time, I modified the example that builds the payloads to remove the wait() calls, which if I understand correctly, shouldn't be used in production (not that my thing is still production, but trying to get there at some point). Maybe the invalid auth key is about something I did wrong there? In any case, I thought the new format could be an improvement from the version with the waits.

sabeatra:
`public func post(_ descriptor: Descriptor, completion: @escaping (Error?) -> Void) throws {
    let verbose = true
    
    let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
    let signer = try APNSSigner(filePath: Keys.apnAuthKeyePath)
    let apnsConfig = APNSConfiguration(
        keyIdentifier: Keys.apnKey,
        teamIdentifier: Keys.teamID,
        signer: signer,
        topic: descriptor.isDriver ? Keys.driverBundleID : Keys.passengerBundleID,
        environment: .sandbox
    )
    
    var connection: APNSConnection?
    
    APNSConnection.connect(configuration: apnsConfig, on: group.next())
    .flatMap { apns -> EventLoopFuture<Void> in
        if verbose {
            print("* Connected to \(apnsConfig.url.host!) (\(apns.channel.remoteAddress!)")
        }
        
        let alert = APSPayload.APSAlert(title: descriptor.title, subtitle: descriptor.subtitle, body: descriptor.body)
        let apsSound = APSPayload.APNSSoundDictionary(
            isCritical: descriptor.isCriticalSound == true,
            name: descriptor.soundName ?? "",
            volume: descriptor.soundVolume ?? 0.0
        )
        let aps = APSPayload(
            alert: alert,
            badge: 0,
            sound: descriptor.isCriticalSound == true ? .critical(apsSound) : .normal(apsSound.name),
            hasContentAvailable: true
        )
        let notification = AcmeNotification(acme2: ["bang", "whiz"], aps: aps)
        
        connection = apns
        return apns.send(
            notification,
            to: descriptor.deviceToken,
            with: CustomEncoder(),
            expiration: descriptor.expirationDate,
            priority: descriptor.priority,
            collapseIdentifier: descriptor.collapseIdentifier
        )
    }.always {
        switch $0 {
        case .success:
            completion(nil)
        case .failure(let error):
            completion(error)
        }
    }.flatMap {
        return connection!.close()
    }.whenComplete { _ in
        group.shutdownGracefully(queue: DispatchQueue.global()) { if let error = $0 { print(error) } }
    }
}`

kylebrowning:
Yes example code is not meant to be used in production. But ill leave this here for posterity!

PR #35, opened 2019-07-23
Fixes #34  [BUG] TooManyProviderTokenUpdates

### LINKED ISSUE #34: [BUG] TooManyProviderTokenUpdates
@kylebrowning Thanks for putting so much work in on this tool! I've been attempting to implement it in a project I'm working on and I noticed the following:

When sending an APNS message to multiple tokens, each time APNSwiftRequestEncoder.write() is called the jwt is regenerated. This causes Apple's servers to respond with **TooManyProviderTokenUpdates**

**To Reproduce**
Configure a connection to Apple's APNS service.
Send more than 5 messages to different tokens.
The service will start to reply with **TooManyProviderTokenUpdates**

After dumping the object to the terminal I noticed that the bearer tokens were different for each requested message. They bearer token should have been the same for each message.

```swift
HeaderTableEntry(name: "authorization", value: "bearer **eyJ[redatcted]ifQ.eyJpc[redatcted]cxfQ.[redacted]490i**") HeaderTableEntry(name: ":path", value: "/3/device/_redacted_________________________1325e345a0daef141dec8c0bd78d") 

HeaderTableEntry(name: "authorization", value: "bearer **eyJ[redatcted]ifQ.eyJpc[redatcted]cwfQ.[redacted]RjTg**") HeaderTableEntry(name: ":path", value: "/3/device/________________________redacted_________________________005e320") 

HeaderTableEntry(name: "authorization", value: "bearer **eyJ[redatcted]ifQ.eyJpc[redatcted]cwfQ.[redacted]od1k**") HeaderTableEntry(name: ":path", value: "/3/device/27a18________________________redacted_________________________9cc1b") 
```

**Expected behavior**
jwt's should only be generated at most once every 20 minutes. There should be some sort of state/expiration on JWT generation so that the target does not throw **TooManyProviderTokenUpdates**

**Platform:**
Ubuntu 18.04
Linux swiftynix 4.15.0-51-generic #55-Ubuntu SMP Wed May 15 14:27:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

itcohorts:
This PR can be used as a reference for fixing this bug. I've tested in my own private project and it works as expected. Hope this helps.

### DIFF
--- a/Sources/APNSwift/APNSwiftBearerToken.swift
+++ b/Sources/APNSwift/APNSwiftBearerToken.swift
@@ -1,0 +1,70 @@
+//
+//  APNSwiftBearerToken.swift
+//  APNSwift
+//
+//  Created by Benjamin Ingmire on 7/23/19.
+//
+
+import Foundation
+
+public class APNSwiftBearerToken
+{
+    let configuration: APNSwiftConfiguration
+    let timeout: TimeInterval
+    var cachedToken : (TimeInterval, String) = (Date().timeIntervalSince1970, "") //JWT found in position 1
+    
+    public init(configuration: APNSwiftConfiguration, timeout: TimeInterval)
+    {
+        self.configuration = configuration
+        self.timeout = timeout
+    }
+    
+    func token() -> String
+    {
+        let getToken = getFreshestToken()
+        return getToken.1
+    }
+    
+    func getFreshestToken()-> (TimeInterval, String)
+    {
+        if self.cachedToken.1 == ""
+        {
+            self.cachedToken = (Date().timeIntervalSince1970, self.createToken())
+        }
+        else
+        {
+            let now = Date().timeIntervalSince1970
+            let diff = now - self.cachedToken.0
+            if diff >= timeout
+            {
+                //refresh the token
+                self.cachedToken = (Date().timeIntervalSince1970, self.createToken())
+            }
+        }
+        return self.cachedToken
+    }
+    
+    func createToken() -> String
+    {
+        let jwt = APNSwiftJWT(keyID: configuration.keyIdentifier, teamID: configuration.teamIdentifier, issueDate: Date(), expireDuration: timeout)
+        var token: String
+        do {
+            let digestValues = try jwt.getDigest()
+            let signature = try configuration.signer.sign(digest: digestValues.fixedDigest)
+            guard let data = signature.getData(at: 0, length: signature.readableBytes) else {
+                throw APNSwiftError.SigningError.invalidSignatureData
+            }
+            token = digestValues.digest + "." + data.base64EncodedURLString()
+        } catch {
+            var ErrorStack = String()
+            Thread.callStackSymbols.forEach {
+                ErrorStack = "\(ErrorStack)\n" + $0
+            }
+            print(ErrorStack)
+            token = ""
+        }
+        return token
+    }
+}
+
+
--- a/Sources/APNSwift/APNSwiftBearerToken.swift
+++ b/Sources/APNSwift/APNSwiftBearerToken.swift
@@ -25,14 +25,16 @@ public struct APNSwiftBearerToken {
         self.timeout = timeout
     }
     
-    public mutating func token() -> String? {
-        let now = Date().timeIntervalSince1970
-        guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else {
-            cachedToken = try? createToken()
-            createdAt = now
-            return cachedToken
+    public var token: String? {
+        mutating get {
+            let now = Date().timeIntervalSince1970
+            guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else {
+                cachedToken = try? createToken()
+                createdAt = now
+                return cachedToken
+            }
+            return existingToken
         }
-        return existingToken
     }
     
     private func createToken() throws -> String {
--- a/Sources/APNSwift/APNSwiftBearerToken.swift
+++ b/Sources/APNSwift/APNSwiftBearerToken.swift
@@ -1,70 +1,49 @@
+//===----------------------------------------------------------------------===//
 //
-//  APNSwiftBearerToken.swift
-//  APNSwift
+// This source file is part of the APNSwift open source project
 //
-//  Created by Benjamin Ingmire on 7/23/19.
+// Copyright (c) 2019 the APNSwift project authors
+// Licensed under Apache License v2.0
 //
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of APNSwift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
 
 import Foundation
 
-public class APNSwiftBearerToken
-{
+public struct APNSwiftBearerToken {
     let configuration: APNSwiftConfiguration
     let timeout: TimeInterval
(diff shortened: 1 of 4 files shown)


Sources/APNSwift/APNSwiftBearerToken.swift
```diff
@@ -0,0 +1,67 @@
...
+        self.timeout = timeout
+    }
+    
+    func token() -> String {
+        let getToken = getFreshestToken()
+        return getToken.1
+    }
+    
+    func getFreshestToken()-> (TimeInterval, String) {
+        if self.cachedToken.1 == "" {
```
kylebrowning: Why not use optionals here and named tuples??

```swift
var cachedToken : (timeout: TimeInterval, token: String)?
guard let existingToken = cachedToken, let now = Date().timeIntervalSince1970, diff <= timeout else { 
// create new Token
return
}
return cachedToken
```

itcohorts: Thanks, I'm pretty new to swift so learning as I go. Appreciate the feedback.

kylebrowning: No worries. Was just curious if I was missing something! Happy to help more, but won't be able to get to this until probably tomorrow.

kylebrowning:
@itcohorts Im wondering If this is cleaner.

```swift
public class APNSwiftBearerToken {
    let configuration: APNSwiftConfiguration
    let timeout: TimeInterval
    var createdAt: TimeInterval?
    private var cachedToken: String?
    
    public init(configuration: APNSwiftConfiguration, timeout: TimeInterval) {
        self.configuration = configuration
        self.timeout = timeout
    }
    
    public var token: String? {
        let now = Date().timeIntervalSince1970
        guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else {
            cachedToken = try? createToken()
            return cachedToken
        }
        return existingToken
    }
    
    private func createToken() throws -> String {
        let jwt = APNSwiftJWT(keyID: configuration.keyIdentifier, teamID: configuration.teamIdentifier, issueDate: Date(), expireDuration: timeout)
        var token: String
        let digestValues = try jwt.getDigest()
        let signature = try configuration.signer.sign(digest: digestValues.fixedDigest)
        guard let data = signature.getData(at: 0, length: signature.readableBytes) else {
            throw APNSwiftError.SigningError.invalidSignatureData
        }
        token = digestValues.digest + "." + data.base64EncodedURLString()
        createdAt = Date().timeIntervalSince1970
        return token
    }
}
```

Then in `APNSwiftRequestEncoder.swift`

```swift
        guard let token = bearerToken.token else {
            promise?.fail(APNSwiftError.SigningError.invalidSignatureData)
            return
        }
        reqHead.headers.add(name: "authorization", value: "bearer \(token)")
```

Sources/APNSwift/APNSwiftBearerToken.swift
```diff
@@ -0,0 +1,61 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+public class APNSwiftBearerToken {
+    let configuration: APNSwiftConfiguration
+    let timeout: TimeInterval
+    var cachedToken: (timeCreated: TimeInterval, token: String)?
```
weissi: this class isn't thread-safe, probably it should be, no?

kylebrowning: Yeah its should be moved to a `struct`. no point in having a mutable class.

itcohorts:
Would moving to a struct prevent the APNSwiftBearerToken from yielding a new token when it timeout is exceeded? Perhaps a manger that yields tokens is needed instead? While a mutable struct is possible, my understanding is that it would have the same thread safety concerns. Appreciate all the feedback!

kylebrowning:
It's my understanding that structs are thread safe because they are value types, not reference types.

itcohorts:
Agreed but my question arises since `public var token: String? {` needs to mutate the struct. My research indicates that the same issue of thread safety arises with a mutating struct. Please correct me if I'm wrong.

itcohorts:
Implemented as a struct. Hoping to be closer with the latest commit. Thanks!

kylebrowning:
Thanks. I’m just double checking the safety!  Will most likely merge by Friday. 

Do let me know if see any other issues and when this goes live in production for you! 

Happy Swifting

Sources/APNSwift/APNSwiftBearerToken.swift
```diff
@@ -0,0 +1,60 @@
...
+        var token: String
+        do {
+            let digestValues = try jwt.getDigest()
+            let signature = try configuration.signer.sign(digest: digestValues.fixedDigest)
+            guard let data = signature.getData(at: 0, length: signature.readableBytes) else {
+                throw APNSwiftError.SigningError.invalidSignatureData
+            }
+            token = digestValues.digest + "." + data.base64EncodedURLString()
+        } catch {
+            var errorStack = String()
```
weissi: please don't print the stack trace here. why don't you make `createToken()` throwing? The user needs to handle the error if there's one, the library shoudn't just return the empty string.

kylebrowning: I’ve resolved that here. 

https://github.com/kylebrowning/APNSwift/pull/35#issuecomment-514433806

kylebrowning:
@itcohorts Can you do something more like this, 
https://github.com/kylebrowning/APNSwift/pull/35#issuecomment-514433806

Specifically, the error stack is an issue, and we should just throw.

Id also prefer `func token()` become a computer property, `var token: String? { get }`

kylebrowning:
@itcohorts Im happy to do, just would love to get you some contribution credit in the git history :)

itcohorts:
Sorry, was a bit out of band taking care of some other things. I'm on it
right now. ;)
Thanks!

itcohorts:
@kylebrowning I addressed the stack issue and made sure that createToken() throws. Thanks for your patience. Regarding moving representing token as a computed property: I tried that as I agree it looks more clean. However I was unable to compile as I kept getting this error: `error: cannot use mutating member on immutable value: 'self' is immutable` I also learned that you cannot mark a computed property as mutating since I tried that too. Please let me know if more changes are required. I appreciate your insights and patience.

Sources/APNSwift/APNSwiftBearerToken.swift
```diff
@@ -0,0 +1,49 @@
...
+        }
+        return existingToken
+    }
+    
+    private func createToken() throws -> String {
+        let jwt = APNSwiftJWT(keyID: configuration.keyIdentifier, teamID: configuration.teamIdentifier, issueDate: Date(), expireDuration: timeout)
+        var token: String
+        let digestValues = try jwt.getDigest()
+        let signature = try configuration.signer.sign(digest: digestValues.fixedDigest)
+        guard let data = signature.getData(at: 0, length: signature.readableBytes) else {
```
weissi: That isn’t quite right, this should be

    signature.readData(length: signature.readableBytes)

And `var signature =` instead of `let`

itcohorts: Thank you, updated.

kylebrowning:
@itcohorts You basically need to set mutating on the get.

```swift
    public var token: String? {
        mutating get {
            let now = Date().timeIntervalSince1970
            guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else {
                cachedToken = try? createToken()
                return cachedToken
            }
            return existingToken
        }
    }
```

PR #38, opened 2019-08-09
Add version shim over `ECDSA_SIG_get0`

t089:
#36

Hey, this would be my suggestion for the version shim as @Lukasa suggested.

Not sure if it is cool to put this directly in `shim.h`.

Also not sure if my use of the `OpaquePointer` tricks is correct (CC @Lukasa).

I've added a test for the signing and ran it on Linux (`swift:5.0-xenial` and `swift:5.0`), seems to be fine 🤞

### DIFF
--- a/Sources/APNSwift/APNSwiftSigner.swift
+++ b/Sources/APNSwift/APNSwiftSigner.swift
@@ -59,15 +59,16 @@ public struct APNSwiftSigner {
         // as this method is `get0` there is no requirement to free those pointers: ECDSA_SIG will free them for us.
         CAPNSOpenSSL_ECDSA_SIG_get0(.init(sig), &r, &s)
         
-        var rb = [UInt8](repeating: 0, count: Int(BN_num_bits(r)+7)/8)
-        var sb = [UInt8](repeating: 0, count: Int(BN_num_bits(s)+7)/8)
-        let lenr = Int(BN_bn2bin(r, &rb))
-        let lens = Int(BN_bn2bin(s, &sb))
-
-        let finalSig = Array(rb[0..<lenr] + sb[0..<lens])
+        var byteBuffer = ByteBufferAllocator().buffer(capacity: Int(BN_num_bits(r)+7)/8 + Int(BN_num_bits(s)+7)/8)
         
-        var byteBuffer = ByteBufferAllocator().buffer(capacity: finalSig.count)
-        byteBuffer.writeBytes(finalSig)
+        byteBuffer.writeWithUnsafeMutableBytes { (p) -> Int in
+            return Int(BN_bn2bin(r, p.baseAddress?.assumingMemoryBound(to: UInt8.self)))
+        }
+        
+        byteBuffer.writeWithUnsafeMutableBytes { (p) -> Int in
+            return Int(BN_bn2bin(s, p.baseAddress?.assumingMemoryBound(to: UInt8.self)))
+        }
+        
         return byteBuffer
     }
 }
--- a/Sources/APNSwift/APNSwiftSigner.swift
+++ b/Sources/APNSwift/APNSwiftSigner.swift
@@ -45,16 +45,22 @@ public struct APNSwiftSigner {
         }
         defer { EC_KEY_free(opaquePointer) }
         
-        let sig = try digest.withUnsafeReadableBytes { ptr -> UnsafeMutablePointer<ECDSA_SIG> in
+        let sig = try digest.withUnsafeReadableBytes { ptr -> OpaquePointer in
             guard let sig = ECDSA_do_sign(ptr.baseAddress?.assumingMemoryBound(to: UInt8.self), Int32(digest.readableBytes), opaquePointer) else {
                 throw APNSwiftError.SigningError.invalidSignatureData
             }
-            return sig
+            return .init(sig)
         }
-        defer { ECDSA_SIG_free(sig) }
+        defer { ECDSA_SIG_free(.init(sig)) }
 
-        let r = sig.pointee.r
-        let s = sig.pointee.s
+        let pr = UnsafeMutablePointer<UnsafePointer<BIGNUM>?>.allocate(capacity: 1)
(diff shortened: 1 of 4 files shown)


Sources/APNSwift/APNSwiftSigner.swift
```diff
@@ -45,16 +45,22 @@ public struct APNSwiftSigner {
...
             }
-            return sig
+            return .init(sig)
         }
-        defer { ECDSA_SIG_free(sig) }
+        defer { ECDSA_SIG_free(.init(sig)) }
 
-        let r = sig.pointee.r
-        let s = sig.pointee.s
+        let pr = UnsafeMutablePointer<UnsafePointer<BIGNUM>?>.allocate(capacity: 1)
```
weissi: where are these two free'd?

weissi:
CC @lukasa

t089:
@weissi this Pointer stuff always confuses me. But yes I think that is missing. The bignums will be freed when the ECDSA_SIG is freed but I would still need to deallocate the pointers to the bignums correct?

Sources/APNSwift/APNSwiftSigner.swift
Lukasa: Rather than ask that question, let's not allocate the pointers at all.

```swift
var pr: UnsafePointer<BIGNUM>? = nil
var ps: UnsafePointer<BIGNUM>? = nil

CAPNSOpenSSL_ECDSA_SIG_get0(.init(sig), &pr, &ps)
```

Then we don't have to free those pointers.

Lukasa: Also, did you actually try to compile this with OpenSSL 1.1? I'm pretty sure `BIGNUM` is also opaque in OpenSSL 1.1, which means you can't actually refer to `BIGNUM` by name.

t089: Yes, I compiled and ran the test in `swift:5.0` which installs openssl-1.1.1

```
docker run swift:5.0 /bin/bash -c "apt-get -q update &&     apt-get -q install -y     openssl libssl-dev && openssl version"
...
OpenSSL 1.1.1  11 Sep 2018
```

We could still use opaque pointe to be on the safe side? How would you declare `pr` and `ps` in that case?

t089: I've updated the PR.

t089: Also, as a side question: I don't quite understand why we can't still use the `ECDSA_SIG` symbol. Isn't it still in the [header](https://github.com/openssl/openssl/blob/59d846ffb13d36d8a8caad09f43bcc968b21cf74/include/openssl/ec.h#L1125)?

Sources/APNSwift/APNSwiftSigner.swift
```diff
@@ -45,26 +45,29 @@ public struct APNSwiftSigner {
...
+            return .init(sig)
         }
-        defer { ECDSA_SIG_free(sig) }
-
-        let r = sig.pointee.r
-        let s = sig.pointee.s
+        defer { ECDSA_SIG_free(.init(sig)) }
 
+        var r : UnsafePointer<BIGNUM>? = nil
+        var s : UnsafePointer<BIGNUM>? = nil
```
Lukasa: I'll be honest: I have no idea why this compiles. As far as I can see in the header files this type should be uninhabited and so we shouldn't be able to declare a pointer to it, but somehow this is all type checking appropriately so I think I'm inclined to leave it alone. As we are never actually dereferencing this pointer nothing bad is going to happen, so we can probably leave it.

Sources/APNSwift/APNSwiftSigner.swift
```diff
@@ -45,26 +45,29 @@ public struct APNSwiftSigner {
...
+        
+        // as this method is `get0` there is no requirement to free those pointers: ECDSA_SIG will free them for us.
+        CAPNSOpenSSL_ECDSA_SIG_get0(.init(sig), &r, &s)
+        
         var rb = [UInt8](repeating: 0, count: Int(BN_num_bits(r)+7)/8)
         var sb = [UInt8](repeating: 0, count: Int(BN_num_bits(s)+7)/8)
         let lenr = Int(BN_bn2bin(r, &rb))
         let lens = Int(BN_bn2bin(s, &sb))
 
         let finalSig = Array(rb[0..<lenr] + sb[0..<lens])
```
Lukasa: Are we interested in throwing away this interim Array? We don't need it, we can simply write in the buffers we want directly to the `byteBuffer`.

t089: I now use `byteBuffer.writeWithUnsafeMutableBytes` to write the bignum bytes directly into the byte buffer, wdyt?

Lukasa: Sorry, I think it's better for you to undo half of that: the two `rb` and `sb` arrays are worth keeping.

`writeWithUnsafeMutableBytes` is a surprising function, in a bad way. In particular, you may have not got enough bytes to write in the data you want, as you can't hint to NIO how many bytes you actually want to receive.

A better version of this API would be to provide an extra length field, but we don't have it for now. So I recommend using the `rb` and `sb` Arrays for now, and then using `ByteBuffer.writeBytes` to write those two buffers in directly.

kylebrowning:
```swift
var signatureBytes = ByteBufferAllocator().buffer(capacity: lenr + lens)
signatureBytes.writeBytes(rb[0..<lenr])
signatureBytes.writeBytes(sb[0..<lens])
return signatureBytes
```

Sources/APNSwift/APNSwiftSigner.swift
t089: Roger that. Will do that then.

t089:
Maybe somebody else could double check that it actually compiles with OpenSSL 1.1.1 :)

kylebrowning:
@t089 Im going to merge this into this branch so the tests run, then ill double check compilation on 1.1

PR #42, opened 2019-09-11
Fix toManyProviderTokenUpdates error

MaciejGad:
I believe that in the current implementation you can still have `toManyProviderTokenUpdates` error. It is because `APNSwiftBearerToken` is a structure and `APNSwiftRequestEncoder` always get own copy of  `APNSwiftBearerToken`, so when  `APNSwiftBearerToken` store its `cachedToken` it is available only in this exact copy. When you call `send<Notification>(...)` you create a new `APNSwiftRequestEncoder` with a new copy of `APNSwiftBearerToken`, that don't have  `cachedToken`. The easiest way to fix it is to change `APNSwiftBearerToken` to class, so it doesn't have copies and when you set `cachedToken` it stays.

### DIFF
--- a/Sources/APNSwift/APNSwiftBearerToken.swift
+++ b/Sources/APNSwift/APNSwiftBearerToken.swift
@@ -14,7 +14,7 @@
 
 import Foundation
 
-public struct APNSwiftBearerToken {
+public class APNSwiftBearerToken {
     let configuration: APNSwiftConfiguration
     let timeout: TimeInterval
     var createdAt: TimeInterval?
@@ -26,7 +26,7 @@ public struct APNSwiftBearerToken {
     }
     
     public var token: String? {
-        mutating get {
+        get {
             let now = Date().timeIntervalSince1970
             guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else {
                 cachedToken = try? createToken()
--- a/Sources/APNSwift/APNSwiftRequestEncoder.swift
+++ b/Sources/APNSwift/APNSwiftRequestEncoder.swift
@@ -27,7 +27,7 @@ internal final class APNSwiftRequestEncoder<Notification>: ChannelOutboundHandle
     typealias OutboundOut = HTTPClientRequestPart
 
     let configuration: APNSwiftConfiguration
-    var bearerToken: APNSwiftBearerToken
+    let bearerToken: APNSwiftBearerToken
     let deviceToken: String
     let priority: Int?
     let expiration: Date?


kylebrowning:
Good callout thanks.

itcohorts:
I have the code working as implemented in a production instance at the moment. I can provide some implementation assistance if you like. Out of office right now but returning in a few hours.

itcohorts:
Sorry for the late reply.

@MaciejGad has a clean fix. I'm not opposed to it, but I was able to get the code to work as implemented. Of course, now looking at it more closely, I'm doing what he has done but at the implementation level instead by storing a mutable copy of `APNSwiftBearerToken`.

In my case I was able to make the struct code work by utilizing an app level connection pool (which is needed for my app.) The object is below, but essentially it holds a copy of connection and a copy of the `bearerToken`. When requesting `APNSwiftBearerToken.token` from `myAPNSConnection` the token is refreshed if the time limit has expired. 

This approach was also very useful as the connection to APNS will eventually time out and a `ChannelError` will be thrown. When that happens it is trivial to catch the `ChannelError` and reconnect the forwarders.

`
public class myAPNSConnection
{
    let type : String
    let connection : APNSwiftConnection
    public var bearerToken : APNSwiftBearerToken //mutable as a copy
    
    public init(type: String, connection: APNSwiftConnection, bearerToken : APNSwiftBearerToken)
    {
        self.type = type
        self.connection = connection
        self.bearerToken = bearerToken
    }
}
`

kylebrowning:
@weissi Im not sure we can add a test that mimics this behavior. It's a response from Apple and unless we know exactly how Apple rejects for too many token provider updates, were gonna be taking shots in the dark.

Apple documentation states
```
The provider’s authentication token is being updated too often. Update the authentication token no more than once every 20 minutes.
```

weissi:
@kylebrowning `EmbeddedChannel` uses `EmbeddedEventLoop` where you can set for the `EventLoop` so if you have tasks scheduled on the `EventLoop`, you can make them fire deterministically. You can find an example here: https://github.com/apple/swift-nio/blob/master/Tests/NIOTests/EmbeddedEventLoopTest.swift#L194-L204

kylebrowning:
@itcohorts maybe you'd be interested in submitting a PR that catches those connection drops too?

kylebrowning:
@weissi That doesn't seem to work with TimeInterval's.

```swift
        let loop = EmbeddedEventLoop()
        var createdToken: String?
        var bearerToken = APNSwiftBearerToken(configuration: apnsConfig, timeout: 10.0))
        createdToken = bearerToken.token
        let task = loop.scheduleTask(in: .seconds(20)) { bearerToken.token!}
        task.futureResult.whenSuccess { createdToken = $0 }

        loop.advanceTime(by: .seconds(9))
        
        XCTAssertTrue(createdToken == bearerToken.token)
        
        loop.advanceTime(by: .seconds(21))
        XCTAssertFalse(createdToken == bearerToken.token)
```

We check the current time in the create token method by 

```swift
let now = Date().timeIntervalSince1970
guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else
```

kylebrowning:
Closing in favor of #49

kylebrowning:
@MaciejGad @itcohorts We've deprecated a few things and put the token under the hood, as well as added tests. Can you double check `master` works for you now?

PR #44, opened 2019-09-20
Linux bignum fix

### LINKED ISSUE #41: Build fails on Linux : use of undeclared type 'BIGNUM'
Building on Linux fails with several of the following

```
/srv/datadisk01/www/VideoSaverServer/.build/checkouts/APNSwift/Sources/APNSwift/APNSwiftSigner.swift:56:31: error: use of undeclared type 'BIGNUM'
        var r : UnsafePointer<BIGNUM>? = nil
```

Swift version 5.1-dev
Ubuntu 18

kylebrowning:
Attempts to fix #41


Sources/APNSwift/APNSwiftSigner.swift
```diff
@@ -40,29 +40,34 @@ public struct APNSwiftSigner {
...
-        var r : UnsafePointer<BIGNUM>? = nil
-        var s : UnsafePointer<BIGNUM>? = nil
+        var r : OpaquePointer? = nil
+        var s : OpaquePointer? = nil
         
         // as this method is `get0` there is no requirement to free those pointers: ECDSA_SIG will free them for us.
-        CAPNSOpenSSL_ECDSA_SIG_get0(.init(sig), &r, &s)
+        withUnsafeMutablePointer(to: &r) { rPtr in
+            withUnsafeMutablePointer(to: &s) { sPtr in
+                CAPNSOpenSSL_ECDSA_SIG_get0(.init(sig), .make(optional: rPtr), .make(optional: sPtr))
```
weissi: I think you can drop both `.make(optional: ...)` here, just `rPtr` and `sPtr` should work

weissi:
CC @lukasa

Sources/APNSwift/APNSwiftSigner.swift
kylebrowning: When I did that I get the following.

Cannot convert value of type 'UnsafeMutablePointer<_>' to expected argument type 'UnsafeMutablePointer<UnsafePointer<BIGNUM>?>?' (aka 'Optional<UnsafeMutablePointer<Optional<UnsafePointer<bignum_st>>>>')

weissi: oh yes, ignore me.

PR #46, opened 2019-09-21
Adds apns-push-type for iOS 13

kylebrowning:


Sources/APNSwift/APNSwiftConnection.swift
```diff
@@ -94,7 +94,7 @@ public final class APNSwiftConnection {
      try apns.send(notification, bearerToken: bearerToken,to: "b27a07be2092c7fbb02ab5f62f3135c615e18acc0ddf39a30ffde34d41665276", with: JSONEncoder(), expiration: expiry, priority: 10, collapseIdentifier: "huro2").wait()
      ```
      */
-    public func send<Notification: APNSwiftNotification>(_ notification: Notification, bearerToken: APNSwiftBearerToken, to deviceToken: String, with encoder: JSONEncoder = JSONEncoder(), expiration: Date? = nil, priority: Int? = nil, collapseIdentifier: String? = nil, topic: String? = nil, pushType: APNSwiftConnection.PushType? = nil) -> EventLoopFuture<Void> {
```
weissi: this is SemVer major breaking

weissi: this is SemVer major breaking

kylebrowning: hmm, how? we added a new send, with the exact signature of the old?

kylebrowning: https://github.com/kylebrowning/APNSwift/pull/46/files#diff-b2725ac51df2d9702582318f5b9449bdR126

PR #48, opened 2019-09-23
Correctly generate signature.

### LINKED ISSUE #39: flaky test
**Describe the bug**
there's a flaky test:

```
Test Case '-[APNSwiftJWTTests.JWTTests testJWTSigning]' started.
/private/tmp/.nio-release-tools_nHItWh/APNSwift/Tests/APNSwiftJWTTests/JWTTests.swift:77: error: -[APNSwiftJWTTests.JWTTests testJWTSigning] : XCTAssertEqual failed: ("63") is not equal to ("64")
Test Case '-[APNSwiftJWTTests.JWTTests testJWTSigning]' failed (0.288 seconds).
```

[log_swift_wmu81k.log](https://github.com/kylebrowning/APNSwift/files/3521659/log_swift_wmu81k.log)

**To Reproduce**

run `swift test --sanitize=address -c release -Xswiftc -enable-testing --disable-index-store` many times

**Expected behavior**

tests pass

**Platform:**
macOS 10.15 beta 6ish

kylebrowning:
fixes #39

### DIFF
--- a/Sources/APNSwift/APNSwiftSigner.swift
+++ b/Sources/APNSwift/APNSwiftSigner.swift
@@ -70,8 +70,9 @@ public struct APNSwiftSigner {
         let lens = Int(BN_bn2bin(.make(optional: s), &sb))
 
         var signatureBytes = ByteBufferAllocator().buffer(capacity: lenr + lens)
-        signatureBytes.writeBytes(rb[0..<lenr])
-        signatureBytes.writeBytes(sb[0..<lens])
+        let allZeroes = Array(repeating: UInt8(0), count: 32)
+        signatureBytes.writeBytes([allZeroes, rb].joined().suffix(32))
+        signatureBytes.writeBytes([allZeroes, sb].joined().suffix(32))
         return signatureBytes
     }
 }


Sources/APNSwift/APNSwiftSigner.swift
```diff
@@ -70,8 +70,9 @@ public struct APNSwiftSigner {
         let lens = Int(BN_bn2bin(.make(optional: s), &sb))
 
         var signatureBytes = ByteBufferAllocator().buffer(capacity: lenr + lens)
-        signatureBytes.writeBytes(rb[0..<lenr])
-        signatureBytes.writeBytes(sb[0..<lens])
+        let allZeroes = Array(repeating: UInt8(0), count: 32)
+        signatureBytes.writeBytes([allZeroes, rb].joined().suffix(32))
+        signatureBytes.writeBytes([allZeroes, sb].joined().suffix(32))
```
t089: Can you explain to me why this change is necessary? Do I understand this correctly that you prepend 32 zero bytes to the actual bytes only to cut off the last 32 bytes again?!

kylebrowning: Did something break for you? This was needed to fix a flaky test.

t089: No, nothing broke. It is just our of curiosity, I don't see how `sb[0..<lens]` could have let to a flaky test. And how your changes are supposed to fix this flaky test?

kylebrowning: `lens` is sometimes only 31 so this makes sure its always 32.

t089: Ah, ok, I see, but why is that? Is the signature still correct when the number is only 31 bytes wide?

kylebrowning: I never saw it fail in the wild, nor was I able to make it fail(become 31) when sending push notifications, but I would imagine the signature would be wrong when its 31.

Our test, specifically checks for the length of 32.

t089: Also note that `rb` and `sb` are already always 32 bytes wide. So I wonder what would happen if you instead just use `signatureBytes.writeBytes(rb)`

t089: > Also note that `rb` and `sb` are already always 32 bytes wide. So I wonder what would happen if you instead just use `signatureBytes.writeBytes(rb)`

Ah, wrong. I think I get it now: the return value of `BN_bn2bin` is the number of bytes actually written into the buffer. For small numbers this can be less than `32`. The size of `rb` is calculated by `BN_num_bits(r)+7)/8`. This might also be wrong already for the same reason. So your fix is indeed correct.

There is a method `BN_bn2binpad` which adds the padding automatically. Maybe we should just use this instead? And also init the `rb` and `sb` arrays with a length of `32`.

t089: On second thought: for big-median, should the padding be in the front, or the end of the buffer?

kylebrowning: Im not entirely sure, im new to this particular signature creation. I can say that if you do 
`while swift test --sanitize=address -c release -Xswiftc -enable-testing --disable-index-store; do date; done`

The old way will eventually fail tests, but with this new way we always pass, and always have the correct length.

t089:
I think your patch is correct, but I think even better would be a slightly different approach using the `BN_bn2binpad` API and thus avoiding the need to create so many arrays. I’ll write up a PR.

t089:
> I think your patch is correct, but I think even better would be a slightly different approach using the `BN_bn2binpad` API and thus avoiding the need to create so many arrays. I’ll write up a PR.

Unfortunately, this function is not available on all the different versions of SSL. So, I don't think it is helpful to rework this. Thanks for the explanation, and sorry that I missed this padding issue when proposing this code.

kylebrowning:
@t089 one day we will be able to drop support for openssl 1.0 :)

As an aside, and out of curiosity are you running this project on production?

PR #49, opened 2019-09-23
Initial provider token fixes.

kylebrowning:
fixes #42

### DIFF
--- a/Sources/APNSwift/APNSwiftBearerToken.swift
+++ b/Sources/APNSwift/APNSwiftBearerToken.swift
@@ -13,39 +13,43 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import NIO
 
-public struct APNSwiftBearerToken {
+public class APNSwiftBearerToken {
     let configuration: APNSwiftConfiguration
     let timeout: TimeInterval
     var createdAt: TimeInterval?
+    private var deadline: TimeAmount
+    private var tokenCreatedAt: TimeAmount?
     private var cachedToken: String?
     
-    public init(configuration: APNSwiftConfiguration, timeout: TimeInterval) {
+    public init(configuration: APNSwiftConfiguration, deadline: TimeAmount) {
         self.configuration = configuration
-        self.timeout = timeout
+        self.deadline = deadline
+        self.timeout = TimeInterval(deadline.nanoseconds / 1000000000)
+    }
+    convenience init(configuration: APNSwiftConfiguration, timeout: TimeInterval) {
+        self.init(configuration: configuration, deadline: TimeAmount.seconds(Int64(timeout)))
     }
     
     public var token: String? {
-        mutating get {
-            let now = Date().timeIntervalSince1970
-            guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else {
-                cachedToken = try? createToken()
-                createdAt = now
-                return cachedToken
-            }
-            return existingToken
+        let now = TimeAmount.nanoseconds(Int64(NIODeadline.now().uptimeNanoseconds))
+        guard let existingToken = cachedToken, let timeCreated = tokenCreatedAt, (now - timeCreated) >= deadline else {
+            cachedToken = try? createToken()
+            createdAt = Date().timeIntervalSince1970
+            tokenCreatedAt = now
+            return cachedToken
         }
+        return existingToken
     }
     
     private func createToken() throws -> String {
-        let jwt = APNSwiftJWT(keyID: configuration.keyIdentifier, teamID: configuration.teamIdentifier, issueDate: Date(), expireDuration: timeout)
-        var token: String
+        let jwt = APNSwiftJWT(keyID: configuration.keyIdentifier, teamID: configuration.teamIdentifier, issueDate: Date())
         let digestValues = try jwt.getDigest()
         var signature = try configuration.signer.sign(digest: digestValues.fixedDigest)
         guard let data = signature.readData(length: signature.readableBytes) else {
             throw APNSwiftError.SigningError.invalidSignatureData
         }
-        token = digestValues.digest + "." + data.base64EncodedURLString()
-        return token
+        return digestValues.digest + "." + data.base64EncodedURLString()
     }
 }
--- a/Sources/APNSwift/APNSwiftBearerToken.swift
+++ b/Sources/APNSwift/APNSwiftBearerToken.swift
@@ -15,7 +15,7 @@
 import Foundation
 import NIO
 
-public class APNSwiftBearerToken {
+public struct APNSwiftBearerToken {
     let configuration: APNSwiftConfiguration
     let timeout: TimeInterval
     var createdAt: TimeInterval?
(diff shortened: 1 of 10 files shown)


kylebrowning:
CC @weissi @itcohorts @MaciejGad

Sources/APNSwift/APNSwiftBearerToken.swift
```diff
@@ -13,39 +13,43 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import NIO
 
-public struct APNSwiftBearerToken {
+public class APNSwiftBearerToken {
```
weissi: This breaks public API.

Sources/APNSwift/APNSwiftBearerToken.swift
```diff
@@ -13,39 +13,43 @@
...
-            let now = Date().timeIntervalSince1970
-            guard let existingToken = cachedToken, let timeCreated = createdAt, (now - timeCreated) <= timeout else {
-                cachedToken = try? createToken()
-                createdAt = now
-                return cachedToken
-            }
-            return existingToken
+        let now = TimeAmount.nanoseconds(Int64(NIODeadline.now().uptimeNanoseconds))
+        guard let existingToken = cachedToken, let timeCreated = tokenCreatedAt, (now - timeCreated) >= deadline else {
+            cachedToken = try? createToken()
```
weissi: do you mind adding explicit `self.`s here? It's really hard to review what's writing to a local var and what is writing to `self.`.

I believe the `struct` to `class` change (which we must undo as it's public API breaking) also made this thread-unsafe. And previously it was a type with value semantics and now it's reference semantics.

kylebrowning:
OKay so @weissi Tests currently break. I advancing time forward doesn't actually change `DispatchTime.now()`

So im not really sure how I could validate that it works via tests.

My manual testing worked perfectly though.

Sources/APNSwift/APNSwiftBearerToken.swift
```diff
@@ -13,39 +13,45 @@
...
 public struct APNSwiftBearerToken {
     let configuration: APNSwiftConfiguration
     let timeout: TimeInterval
     var createdAt: TimeInterval?
+    private var deadline: TimeAmount
+    private var tokenCreatedAt: TimeAmount?
     private var cachedToken: String?
     
-    public init(configuration: APNSwiftConfiguration, timeout: TimeInterval) {
+    public init(configuration: APNSwiftConfiguration, deadline: TimeAmount) {
```
weissi: a `TimeAmount` is not a deadline, NIO has `NIODeadline` for that.

Sources/APNSwift/APNSwiftConnection.swift
```diff
@@ -96,8 +95,18 @@ public final class APNSwiftConnection {
...
+                    if configuration.tlsConfiguration.privateKey == nil {
+                        do {
+                            tokenFactory = try APNSwiftBearerTokenFactory(eventLoop: eventLoop, configuration: configuration)
+                        } catch {
+                            return channel.eventLoop.makeFailedFuture(APNSwiftError.SigningError.invalidSignatureData)
+                        }
+                    }
+                    return connectionFullyUpPromise.futureResult.map { () -> APNSwiftConnection in
+                        let connection = APNSwiftConnection(channel: channel, multiplexer: multiplexer, configuration: configuration)
+                        connection.bearerTokenFactory = tokenFactory
```
weissi: couldn't we pass the factory in the initialiser?

kylebrowning: @weissi that's a public init method, so API breaking.

Sources/APNSwift/APNSwiftConnection.swift
```diff
@@ -164,14 +174,18 @@ public final class APNSwiftConnection {
             }
     }
     public func send<Notification: APNSwiftNotification>(_ notification: Notification, bearerToken: APNSwiftBearerToken, to deviceToken: String, with encoder: JSONEncoder = JSONEncoder(), expiration: Date? = nil, priority: Int? = nil, collapseIdentifier: String? = nil, topic: String? = nil) -> EventLoopFuture<Void> {
```
weissi: this one should also be deprecated because of the bearer token I think

Sources/APNSwift/APNSwiftConnection.swift
weissi: @kylebrowning I think we should deprecate all public inits for that and add another (new, additional, `internal`) one that takes the factory. Only changing/deleting the existing one is breaking, adding a new one is not.

So the existing one can just become deprecated and `public convenience init`, no?

kylebrowning: Cant be declared public unless APNSwiftBearerTokenFactory becomes public.

Its also throwing.

weissi: > @weissi that's a public init method, so API breaking.

^^^^^^

weissi: - I would have no public inits at all
- apart from the one that we already have, which I would deprecate
- we can add a new, internal init that the existing one can call into

Sources/APNSwift/APNSwiftConnection.swift
```diff
@@ -118,6 +116,14 @@ public final class APNSwiftConnection {
...
     
+    private init(channel: Channel, multiplexer: HTTP2StreamMultiplexer, configuration: APNSwiftConfiguration, bearerTokenFactory: APNSwiftBearerTokenFactory?) {
+        self.channel = channel
+        self.multiplexer = multiplexer
+        self.configuration = configuration
+        self.bearerTokenFactory = bearerTokenFactory
+    }
+    
+    @available(*, deprecated, message: "APNSwiftConnection is initialized internally now.")
     public init(channel: Channel, multiplexer: HTTP2StreamMultiplexer, configuration: APNSwiftConfiguration) {
```
weissi: this should become a `public convenience init` and call the `private init` from above

Sources/APNSwift/APNSwiftBearerTokenFactory.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  APNSwiftBearerTokenFactory.swift
```
weissi: needs the correct license header

Sources/APNSwift/APNSwiftConnection.swift
```diff
@@ -135,12 +151,12 @@ public final class APNSwiftConnection {
      try apns.send(notification, bearerToken: bearerToken,to: "b27a07be2092c7fbb02ab5f62f3135c615e18acc0ddf39a30ffde34d41665276", with: JSONEncoder(), expiration: expiry, priority: 10, collapseIdentifier: "huro2").wait()
      ```
      */
-    public func send<Notification: APNSwiftNotification>(_ notification: Notification, pushType: APNSwiftConnection.PushType, bearerToken: APNSwiftBearerToken, to deviceToken: String, with encoder: JSONEncoder = JSONEncoder(), expiration: Date? = nil, priority: Int? = nil, collapseIdentifier: String? = nil, topic: String? = nil) -> EventLoopFuture<Void> {
```
weissi: we need to add another (deprecated) `send` method  that has `notification: pushType:bearerToken: to: String, with: expiration: priority: collapseIdentifier: topic:)`

weissi: currently, it's breaking API

weissi: oh, maybe we don't? Because we just introduced `pushType:`?

kylebrowning: @weissi Thats correct, push type is new, not released to public yet.

PR #51, opened 2019-10-02
Support raw payloads.

### LINKED ISSUE #50: Support for raw payloads
I occasionally have a requirement where I have raw string payloads that I'd like to send over APNS. I couldn't see a way to pass a string payload with the current set of classes and protocols. Is this something you'd be willing to support?

E.g. I just want to be able to pass something like the following:

```json
{ "aps": { "alert": "hello world", "badge": 9, "sound": "bingbong.aiff"} ,"g": {"l": "appme:///dispatch/us?utm_medium=notification&utm_source=ec&utm_campaign=20171018"}}
```

kylebrowning:
Fixes #50


daniel-beard:
Looks great!

PR #63, opened 2019-11-25
Add Channel Handler Removed.

### LINKED ISSUE #62: [BUG]Fatal error: leaking promise created at (file: "debug/.build/checkouts/APNSwift/Sources/APNSwift/APNSwiftConnection.swift", line: 177):
I am running a Vapor 4 app on Ubuntu 18.04 with APNSwift 1.3.0.

As of now I just send a simple notification as described in the getting started page of this repo.

I don't know if it is a coincidence but I noticed that the fatal error appears only after a couple of hours without interacting with the service.

Please let me know if I can give you other informations, it is quite a complicated bug to reproduce.

kylebrowning:
This should fix #62 . We were not properly failing promises if any were left in the tank.

### DIFF
--- a/Sources/APNSwift/APNSwiftErrors.swift
+++ b/Sources/APNSwift/APNSwiftErrors.swift
@@ -14,6 +14,9 @@
 
 import Foundation
 
+/// An error for when the connections ends, but we still have promises in our queue.
+public struct NoResponseReceivedBeforeConnectionEnded: Error, Equatable {}
+
 /// This is an enum that provides the possible responses from Apple
 public struct APNSwiftError: Equatable {
     public enum ResponseError: Error, Equatable {
@@ -58,10 +61,6 @@ public struct APNSwiftError: Equatable {
         public var description: String {
             return rawValue
         }
-    }
-    
-    public enum Internal: Error {
-        case handlerRemovedBeforeFullfilled
     }
     
     public enum SigningError: Error {
--- a/Sources/APNSwift/APNSwiftErrors.swift
+++ b/Sources/APNSwift/APNSwiftErrors.swift
@@ -16,7 +16,7 @@ import Foundation
 
 /// An error for when the connections ends, but we still have promises in our queue.
 public struct NoResponseReceivedBeforeConnectionEnded: Error, Equatable {}
-public struct ResponseFromAppleMissing: Error, Equatable {}
+public struct NoResponseBodyFromApple: Error, Equatable {}
 
 /// This is an enum that provides the possible responses from Apple
 public struct APNSwiftError: Equatable {
--- a/Sources/APNSwift/APNSwiftErrors.swift
+++ b/Sources/APNSwift/APNSwiftErrors.swift
@@ -16,6 +16,7 @@ import Foundation
 
 /// An error for when the connections ends, but we still have promises in our queue.
 public struct NoResponseReceivedBeforeConnectionEnded: Error, Equatable {}
+public struct ResponseFromAppleMissing: Error, Equatable {}
 
 /// This is an enum that provides the possible responses from Apple
 public struct APNSwiftError: Equatable {
--- a/Sources/APNSwift/APNSwiftErrors.swift
+++ b/Sources/APNSwift/APNSwiftErrors.swift
@@ -16,6 +16,7 @@ import Foundation
 
 /// An error for when the connections ends, but we still have promises in our queue.
 public struct NoResponseReceivedBeforeConnectionEnded: Error, Equatable {}
+/// An error where a request was made to Apple, but the response body buffer was nil
 public struct NoResponseBodyFromApple: Error, Equatable {}
 
 /// This is an enum that provides the possible responses from Apple
--- a/Sources/APNSwift/APNSwiftErrors.swift
+++ b/Sources/APNSwift/APNSwiftErrors.swift
@@ -59,6 +59,11 @@ public struct APNSwiftError: Equatable {
             return rawValue
         }
     }
+    
+    public enum Internal: Error {
+        case handlerRemovedBeforeFullfilled
+    }
+    
     public enum SigningError: Error {
         case invalidAuthKey
         case invalidASN1
--- a/Sources/APNSwift/APNSwiftStreamHandler.swift
+++ b/Sources/APNSwift/APNSwiftStreamHandler.swift
@@ -46,4 +46,10 @@ final class APNSwiftStreamHandler: ChannelDuplexHandler {
         queue.insert(input, at: 0)
         context.write(wrapOutboundOut(input.request), promise: promise)
     }
+    
+    func handlerRemoved(context: ChannelHandlerContext) {
+        queue.forEach { context in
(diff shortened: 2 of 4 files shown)


Sources/APNSwift/APNSwiftStreamHandler.swift
```diff
@@ -46,4 +46,10 @@ final class APNSwiftStreamHandler: ChannelDuplexHandler {
         queue.insert(input, at: 0)
         context.write(wrapOutboundOut(input.request), promise: promise)
     }
+    
+    func handlerRemoved(context: ChannelHandlerContext) {
+        queue.forEach { context in
```
weissi: please use

```
while let element = queue.popFirst() {
    element.responsePromise.fail(...)
}
```

you can't use `queue.forEach` because fulfilling a promise will call arbitrary code which might _mutate_ queue which would be an exclusivity violation.

Sources/APNSwift/APNSwiftErrors.swift
```diff
@@ -59,6 +59,11 @@ public struct APNSwiftError: Equatable {
             return rawValue
         }
     }
+    
+    public enum Internal: Error {
```
weissi: given that you will never be able to add a case to this `enum` (as that would break public API), you should use

    public struct NoResponseReceivedBeforeConnectionEnded: Error {}

or something like that.

weissi: I also think `handlerRemovedBeforeFulfilled` doesn't mean anything to the user of your library, they probably have no clue what NIO even is.

The fundamental problem here is that you send a request to the server but at the time where the connection went away, you haven't received a response yet from the server.

I should add that it's not necessarily that the connection ended but assuming that your library doesn't remove handlers itself, you can just use `handlerRemoved` as a proxy for connection ended. There's also `channelInactive` which is a more precise event for 'the connection went away'. I still tend to use `handlerRemoved` for these tasks because that will also work in case you do use `pipeline.removeHandler(...)` in which case the error message won't be accurate but everything will still work.

Tests/APNSwiftTests/APNSwiftRequestTests.swift
```diff
@@ -219,10 +219,41 @@ final class APNSwiftRequestTests: XCTestCase {
...
+                let expected = APNSwiftError.Internal.handlerRemovedBeforeFullfilled
+                if error != expected {
+                    XCTFail("response is: \(error), should be: \(expected)")
+                }
+            default:
+                XCTFail("response should not succeed")
+            }
+        }
+        let leftovers = try channel.finish()
+        XCTAssertTrue(leftovers.hasLeftOvers)
```
weissi: if you do `channel.readOutbound` before `.finish()`, then you can use

    XCTAssertNoThrow(XCTAssertTrue(try channel.finish().isClean()))

Sources/APNSwift/APNSwiftStreamHandler.swift
```diff
@@ -36,7 +36,7 @@ final class APNSwiftStreamHandler: ChannelDuplexHandler {
             if var buffer = res.byteBuffer, let data = buffer.readData(length: buffer.readableBytes), let error = try? JSONDecoder().decode(APNSwiftError.ResponseStruct.self, from: data) {
                 return current.responsePromise.fail(APNSwiftError.ResponseError.badRequest(error.reason))
             }
-            return
+            return current.responsePromise.fail(APNSwiftError.ResponseError.badRequest(.unknown))
```
weissi: think this is still using the old erro

kylebrowning: @weissi Thats specifically for the case where `let error = try? JSONDecoder().decode(APNSwiftError.ResponseStruct.self, from: data)` fails.

Should I just return instead like I was before?

kylebrowning: @weissi Does that make sense?

weissi: Apologies, I think I misread the code before. But rereading, why not fail the promise with the actual error from the json decoder?

kylebrowning: @weissi Thats a good point. Okay will do.

Sources/APNSwift/APNSwiftStreamHandler.swift
```diff
@@ -33,10 +33,16 @@ final class APNSwiftStreamHandler: ChannelDuplexHandler {
         let res = unwrapInboundIn(data)
         guard let current = self.queue.popLast() else { return }
         guard res.header.status == .ok else {
-            if var buffer = res.byteBuffer, let data = buffer.readData(length: buffer.readableBytes), let error = try? JSONDecoder().decode(APNSwiftError.ResponseStruct.self, from: data) {
-                return current.responsePromise.fail(APNSwiftError.ResponseError.badRequest(error.reason))
+            if var buffer = res.byteBuffer, let data = buffer.readData(length: buffer.readableBytes) {
```
weissi: you can simplify the whole code here (you don't need `let data = buffer.readData(...)` anymore if you up you `swift-nio` dependency to `2.10.0` because then you can directly decode the codable from a `ByteBuffer`)

```
guard let buffer = res.byteBuffer else {
    return current.responsePromise.fail(ResponseFromAppleMissing())
}
do {
    let error = try JSONDecoder().decode(APNSwiftError.ResponseStruct.self, from: buffer)
    return current.responsePromise.fail(APNSwiftError.ResponseError.badRequest(error.reason))
} catch {
    return current.responsePromise.fail(error)
}
```

weissi: see here for new API: https://apple.github.io/swift-nio/docs/current/NIOFoundationCompat/Extensions/JSONDecoder.html#/s:10Foundation11JSONDecoderC19NIOFoundationCompatE6decode_4fromxxm_3NIO10ByteBufferVtKSeRzlF

Sources/APNSwift/APNSwiftStreamHandler.swift
```diff
@@ -33,10 +33,15 @@ final class APNSwiftStreamHandler: ChannelDuplexHandler {
         let res = unwrapInboundIn(data)
         guard let current = self.queue.popLast() else { return }
         guard res.header.status == .ok else {
-            if var buffer = res.byteBuffer, let data = buffer.readData(length: buffer.readableBytes), let error = try? JSONDecoder().decode(APNSwiftError.ResponseStruct.self, from: data) {
+            guard let buffer = res.byteBuffer else {
+                return current.responsePromise.fail(ResponseFromAppleMissing())
+            }
+            do {
+                let error = try JSONDecoder().decode(APNSwiftError.ResponseStruct.self, from: buffer)
```
weissi: awesome, think you need to change your `Package.swift` to say `from: "2.10.1"` or so too

kylebrowning: Screen Shot 2019-11-26 at 7 43 10 AM


Doesn't appear so, but im happing to do it!

Issue #67, opened 2019-12-03
What should I use for keyIdentifier


grosch:
When doing a push notification based off of a PassKit item (i.e. using a PEM instead of a p8), I'm not sure what I pass for the `keyIdentifier` parameter to the `APNSwiftconfiguration` constructor.

kylebrowning:
you can pass `""` until the convenience initializer is merged.

PR #68, opened 2019-12-03
Provide pem path convenience.

### LINKED ISSUE #66: Add overloads
Seems like quite a bit of boilerplate to generate the configuration.  Would be easier if the initializer, instead of taking a signer parameter, had one method where you passed the path to the p8 file, and a second convenience initializer that instead took the path to the PEM and key files.

kylebrowning:
Fixes #66 and #67

### DIFF
--- a/Sources/APNSwift/APNSwiftConfiguration.swift
+++ b/Sources/APNSwift/APNSwiftConfiguration.swift
@@ -100,7 +100,7 @@ public struct APNSwiftConfiguration {
         ````
         */
     public init(privateKeyPath: String, pemPath: String, topic: String, environment: APNSwiftConfiguration.Environment, logger: Logger? = nil) throws {
-        try self.init(keyIdentifier: "", teamIdentifier: "", signer: APNSwiftSigner(buffer: ByteBufferAllocator().buffer(capacity: Data().count)), topic: topic, environment: environment, logger: logger)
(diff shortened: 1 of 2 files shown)


Sources/APNSwift/APNSwiftConfiguration.swift
```diff
@@ -78,6 +78,34 @@ public struct APNSwiftConfiguration {
...
+                privateKeyPath: "/Users/kylebrowning/Projects/swift/Fern/development_com.grasscove.Fern.pkey",
+                pemPath: "/Users/kylebrowning/Projects/swift/Fern/development_com.grasscove.Fern.pem",
+                topic: "com.grasscove.Fern",
+                environment: .sandbox
+            )
+        )
+        ````
+        */
+    public init(privateKeyPath: String, pemPath: String, topic: String, environment: APNSwiftConfiguration.Environment, logger: Logger? = nil) throws {
+        try self.init(keyIdentifier: "", teamIdentifier: "", signer: APNSwiftSigner(buffer: ByteBufferAllocator().buffer(capacity: Data().count)), topic: topic, environment: environment, logger: logger)
```
weissi: `Data().count` == 0, right? Just put a number there that approximates the real world, maybe 1024 or so?

kylebrowning: Yes that’s correct. It’s not even used in the whole process. It’s only used for the bearer token factory. Which isn’t used in PEM.

PR #72, opened 2019-12-06
Added a convenience method that takes Data

grosch:
It's more likely that somebody is going to have a `Data` object if they're sending raw bytes, so this just hides the underlying byte buffers to make life easier on us.


florianreinhart:
Why not use `DataProtocol` to be even more flexible?

grosch:
You could do that too 😀

weissi:
@grosch if you add `...<Bytes: Collection(..., data: Bytes) where Bytes.Element == UInt8` then it’ll work for Data, DispatchData, Array<UInt8> and many more.

weissi:
(DataProtocol is a sub-protocol of (RandomAccess)Collection)

kylebrowning:
Ive incorporated this into #74

grosch:
@weissi I'm not quite following.  Can you show the exact method signature you're wanting?

kylebrowning:
@grosch Ive already added it. in #74

PR #78, opened 2019-12-20
Allow PEM password

grosch:


Sources/APNSwift/APNSwiftConfiguration.swift
```diff
@@ -78,34 +78,129 @@ public struct APNSwiftConfiguration {
...
+    ///   For all other types of push notifications, please switch to the newer `.p8` file format.
+    ///
+    /// - Parameters:
+    ///   - privateKeyPath: The path to your private key
+    ///   - pemPath: The path to your certificate in PEM format
+    ///   - topic: The bundle identifier for these push notifications
+    ///   - environment: The environment to use:  `sandbox` or `production`
+    ///   - logger: The logger you wish to use.  If `nil`, one will be created
+    ///   - pemPassword: The password to the private key
+    public init(privateKeyPath: String, pemPath: String, topic: String, environment: APNSwiftConfiguration.Environment, logger: Logger? = nil, pemPassword: Data? = nil) throws {
```
weissi: this is API breaking, we need to add a new method (without a default parameter) instead of adding a parameter here.

grosch:
@weissi I split it into two initializers, as requested.

Sources/APNSwift/APNSwiftConfiguration.swift
```diff
@@ -47,8 +47,8 @@ public struct APNSwiftConfiguration {
      signed. Since security is important keeping we extracted this logic into three options.
      `file`, `data`, or `custom`.
        - topic: The bundle identifier for these push notifications.
-       - environment: The environment to use, sandbox, or production.
-       - logger: The logger you wish to use, if nil, one will be created
+       - environment: The environment to use:  `sandbox` or `production`
```
kylebrowning: nit: extra space here.

Sources/APNSwift/APNSwiftConfiguration.swift
```diff
@@ -78,34 +78,148 @@ public struct APNSwiftConfiguration {
...
-                privateKeyPath: "/Users/kylebrowning/Projects/swift/Fern/development_com.grasscove.Fern.pkey",
-                pemPath: "/Users/kylebrowning/Projects/swift/Fern/development_com.grasscove.Fern.pem",
-                topic: "com.grasscove.Fern",
-                environment: .sandbox
-            )
-        )
-        ```
-    */
+
+    /// Creates a new configuration for APNSwift with a PEM key and certificate
```
kylebrowning: Can we use `/*` for comments so its consistent throughout the code base.

It also lookalike you dropped the Usage Example.

grosch: Sure, I can, but you need to use the version then that starts all the lines with a * character, not just leave it blank.  If you don't, as soon as you reindent the file if screws up the method comments.  That's why I like the /// one better.  Think it looks cleaner.

PR #95, opened 2020-12-08
Add source documentation for error responses

florianreinhart:
Descriptions are taken from [the official documentation](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html#//apple_ref/doc/uid/TP40008194-CH11-SW17).


kylebrowning:
Thanks so much @florianreinhart

PR #98, opened 2020-12-16
Add Decode ability to APNSwiftPayload

mackoj:
In order to more easily make it work with [Vapor Queues](https://docs.vapor.codes/4.0/queues/) this allow me to make jobs that have Payload that are the push themselves.

As I asked [here](https://github.com/kylebrowning/APNSwift/discussions/97) I think that making them work inside jobs might be better.


kylebrowning:
thanks @mackoj

PR #101, opened 2021-01-31
Add timeout option if connection does not receive response.

madsodgaard:
This PR provides a work-around for when Apple servers does not return any response, and therefore leaves the connection occupied and blocked. This means in production environments, with 2 connections you are forced to restart the server, if two attempts to send a push notifications hangs because of Apple. This actually happens pretty frequently.

The timeout is disabled by default and can be enabled by setting the `timeout` property of `APNSwiftConfiguration`.

I've tested this in our production environment, and fixes the issues we were having with the connections blocking.

### DIFF
--- a/Sources/APNSwift/APNSwiftConfiguration.swift
+++ b/Sources/APNSwift/APNSwiftConfiguration.swift
@@ -168,6 +168,8 @@ public struct APNSwiftConfiguration {
     public var topic: String
     public var environment: Environment
     internal var logger: Logger?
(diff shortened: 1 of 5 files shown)


kylebrowning:
@madsodgaard thank you so much, this is tremendously helpful.

emland89:
wow thank you. We were having this same issue, restarting the server like every 30 min is not fun, thanks for this. Any ideas as to why apple may not be returning anything?

zastrozzi:
I'd like to give this a +100. Thank you so much.

PR #105, opened 2021-04-30
support to connect via proxy server

SupaBasti:


Package.swift
```diff
@@ -16,6 +16,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.13.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
         .package(url: "https://github.com/vapor/jwt-kit.git", from: "4.0.0"),
+        .package(url: "https://github.com/swift-server/async-http-client", from: "1.2.5"),
```
kylebrowning, line 19: This is a very heavy handed approach that we can do without. I dont see the point of adding a dependency this large when we have all the tools at our disposal with swift-nio-http2

kylebrowning (changes requested):

Package.swift
kylebrowning: For example, we now have two ways of build connections, one with swift-nio and one with AsyncHTTPClient which uses swift nio. we should just use swift-nio

kylebrowning:
https://github.com/kylebrowning/APNSwift/issues/104

kylebrowning:
im closing this in favor of adopting async-http-client for everything.

PR #110, opened 2021-09-07
Update APS Payload Keys for iOS 13

ve6yeq:
Apple has added some extra keys since iOS 13 so update the list of known payload dictionary keys.


kylebrowning:
Thanks!

PR #117, opened 2021-10-20
Add possibility to not log

### LINKED ISSUE #114: Make it possible to not log
Right now, it's not possible to *not* log, when using e.g. [vapor/apns](https://github.com/vapor/apns) (see also [issue #38](https://github.com/vapor/apns/issues/38)), because the `Request.logger` or `Application.logger` is used as a fallback. When the parameter `logger` is set to `nil`, the logging should be disabled.

As a workaround for now, I'm passing the `SwiftLogNoOpLogHandler` (i.e. `Logger(label: "no op logger", factory: { _ in SwiftLogNoOpLogHandler() })`), but I'd rather pass `nil` to remove logging.

I'm happy to open a PR, but maybe we can discuss first about it: At first it's the `self.logger` fallback for the `APNSwiftClient`, but there is another fallback to `self.configuration.logger` in the `APNSwiftConnection`. The second one is not that a huge deal, because we can just pass `nil` in the configuration. The first one could maybe use some kind of default value instead of `nil` to fallback to `self.logger`. What do you think? Or is this behaviour just like intended?

(A small note, why I want to avoid those logs: They are a bit too verbose for me to be logged in log level `info`. I can achieve almost the same logs from "outside" the library and add some more detail to the logs.)

code28:
Fixes #114 

As in #114 already discussed, is this topic not that straight-forward, because there are a lot of logger parameters and properties. I hope, I got the idea, you mentioned in the issue.

I don't really like this solution that much, because it duplicates a lot of code. (We could still add `@available(*, deprecated)` to the old ones..?) I wanted to start with the most obvious solution and am open for better ideas. Without making breaking changes, I think, this still might be the best way.

Unfortunately, the name `.none` for the enum is not the totally correct, because we can still configure a logger in `APNSwiftConnection` as fallback, which can't be excluded at that point.

I'm happy to hear what you think about this!

### DIFF
--- a/Sources/APNSwift/APNSwiftClient.swift
+++ b/Sources/APNSwift/APNSwiftClient.swift
@@ -32,6 +32,16 @@ public protocol APNSwiftClient {
 }
 
 extension APNSwiftClient {
+    private func logger(from loggerConfig: LoggerConfig) -> Logger? {
+        switch loggerConfig {
+        case .none:
+            return nil
+        case .clientLogger:
+            return self.logger
+        case .custom(let customLogger):
+            return customLogger
+        }
+    }
     /**
      APNSwiftConnection send method. Sends a notification to the desired deviceToken.
      - Parameter payload: the alert to send.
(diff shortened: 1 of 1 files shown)


kylebrowning:
Any update here @code28

code28:
> Any update here @code28

Sorry for this long time. Actually, I was waiting for a reply to my comment, but thinking about it, you already made clear, that you'll just release a new major version. So I'm going to update this PR soon!

code28:
I'm sorry that there is still no progress here - it's still on my list, but the weeks before christmas are crazy... 🎄

kylebrowning:
@code28 no rush, I think it will be a fine addition once it's in!

kylebrowning:
thanks @code28, ill get this in, pop off a release, and then I can finish working on async/await

kylebrowning (approved):

PR #124, opened 2021-12-08
[feat] Mark content boolean properties in init as optional

TimAEllis:
Recently I discovered that my background notifications were not being delivered to my application anymore. Upon inspecting the `aps` part of the payload being sent, it would set `"content-available"` and `"mutable-content"` properties. The presence of `"mutable-content"` would cause the payload to not be delivered at all (in my app testing at least).

I have left the initialisers in their original state (of `false` rather than `nil`) in order to not introduce any breaking changes

### DIFF
--- a/Sources/APNSwift/APNSwiftPayload.swift
+++ b/Sources/APNSwift/APNSwiftPayload.swift
@@ -25,12 +25,22 @@ public struct APNSwiftPayload: Codable {
     public let interruptionLevel: String?
     public let relevanceScore: Float?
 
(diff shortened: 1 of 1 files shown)


TimAEllis:
Update to my original comment. Turns out that it wasn't the root cause of the issue. Was me doing a silly and omitting `pushType: .background` in the `send` command, But it still stands that they _could_ be `nil`able to save a few bytes haha

kylebrowning (approved):

PR #142, opened 2022-08-25
Change AuthenticationManager to class with lock

FranzBusch:
# Motivation
Currently, the JWT authentication is using two actors to isolate the state. This leads to thread hops for every APNS request which we can avoid with using a `Lock` instead.

# Modification
This PR contains a couple of modifications to the JWT authentication handling:
- Refactor the two actors to a single class using a `Lock`
- Drop the dependency on `Foundation` from the JWT handling.
- Improve the logging
- Add tests to verify the behaviour of the new JWT generation.

# Result
We now have a faster and better tested JWT handling.


kylebrowning (approved):
This is really smart and a great use of lock. 

I especially like the simplification of the process and the breakout of variables. 

Thank you for doing this! Tests should be running.

PR #143, opened 2022-08-25
Refactor configuration and client

FranzBusch:
# Motivation
This PR is doing multiple things in once and I can't split them apart since the client and the configuration are tightly coupled. The goals of this PR is to create an evolable API that we can use to build more semantically safe constructs. Furthermore, it splits the configuration properly between what can be configured on a single connection and what can be configured on a request.

# Modification
In details this PR does the following things:

1. Every type that is touched in this PR should be forward evolable without API breakage. Therefore, all enums are converted to struct
2. The current configuration mixes configs from the connection and single requests. The new proposed config splits them between the two types.
3. Introduce a new error struct that contains semantic information and more debug information
4. Makes the client generic over the encoder/decoder and introduces a new low level `send` method.

# Result
We now have a new config and a baseline client that allows us to build semantically safe APIs in the next PRs.


Sources/APNSwiftExample/Program.swift
```diff
@@ -0,0 +1,69 @@
...
+            backgroundActivityLogger: logger
+        )
+        defer {
+            client.shutdown { _ in
+                logger.error("Failed to shutdown APNSClient")
+            }
+        }
+
+        do {
+            // TODO: Send pushes once semantic APIs land
```
FranzBusch, line 64: This will be resolved in a follow-up PR

Sources/APNSwift/APNSErrorResponse.swift
```diff
@@ -0,0 +1,30 @@
...
+// See CONTRIBUTORS.txt for the list of APNSwift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A struct for the error response of APNs.
+///
+/// This is just used to decode the JSON and should not be exposed.
+struct APNSErrorResponse: Codable {
```
kylebrowning, line 18: If the comment is correct, should this be `internal`?

Sources/APNSwift/Coding/APNSJSONDecoder.swift
```diff
@@ -0,0 +1,31 @@
...
+/// A protocol that is similar to the ``JSONDecoder``. This allows users of APNSwift to customize the decoder used
+/// for decoding the APNS response bodies.
+public protocol APNSJSONDecoder {
+    func decode<T: Decodable>(_ type: T.Type, from buffer: ByteBuffer) throws -> T
+}
+
+extension JSONDecoder: APNSJSONDecoder {
+    public func decode<T: Decodable>(_ type: T.Type, from buffer: ByteBuffer) throws -> T {
+        var copy = buffer
+        let data = copy.readData(length: buffer.readableBytes)!
```
kylebrowning, line 28: Since this throws, is the force unrwap safe? Perhaps a guard unwrap and then throw if nil?

kylebrowning (approved):

Sources/APNSwift/Coding/APNSJSONDecoder.swift
FranzBusch: Yes this is a safe force-unwrap since it will only return `nil` if there are not `length` readable bytes. However, we are passing `buffer.readableBytes` which guarantees that the length will be readable.

PR #144, opened 2022-08-25
Add semantic safe API for alert notifications

FranzBusch:
# Motivation
We want to provide new APIs that are semantically safe when sending alerts to APNs. These APIs should feel native in Swift and not leak too many APIs of the APNs interface.

# Modification
This PR creates a few new semantic types to represent currency types from the APNs API. Furthermore, the PR adds new types for the alert notification and a convenience method to send alert notifications with the APNSClient.

# Result
We can now send alert notifications.


kylebrowning:
Everything worked and the documentation looks great. Thank you for making this a much cleaner interface.

PR #162, opened 2022-09-21
iOS 16.1 Live Activity support added

nasircsms:


alex-de-waal:
Thank you for this, you are a legend! 

How would I use the Activity's push token in the request's header? As described in the documentation:

> To use remote push notifications to update your Live Activities, you need to use push tokens for authenticating with APNs as described in [Establishing a Token-Based Connection to APNs](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/establishing_a_token-based_connection_to_apns).

> When you’ve successfully started your Live Activity, you receive an [Activity](https://developer.apple.com/documentation/activitykit/activity) with a [pushToken](https://developer.apple.com/documentation/activitykit/activity/pushtoken). Send it to your push notification server, and use it to send remote push notifications that update or end the Live Activity.

> Send push notifications to the Live Activity using the push token you stored on your server. It’s essential that you set the fields for the content-state key to match your custom [Activity.ContentState](https://developer.apple.com/documentation/activitykit/activity/contentstate-swift.typealias) type to ensure the system can decode the JSON payload and update the Live Activity.

nasircsms:
The deviceToken param will contain Activity pushToken. There is no need to send it separately. So we'll be sending push notification using Activity pushToken and deviceToken won't be required in this case. 
I should have updated param name to reflect this.

alex-de-waal:
Awesome, that helps a lot! Thanks again dude

Sources/APNSwift/LiveActivity/APNSClient+LiveActivity.swift
```diff
@@ -0,0 +1,44 @@
...
+//  PushSender
+//
+//  Created by csms on 20/09/2022.
+//
+
+import NIOCore
+import Logging
+
+extension APNSClient {
+    /// Sends an alert notification to APNs.
```
FranzBusch, line 12: This should probably be changed to say sending a live activity notification

Sources/APNSwift/APNSPushType.swift
```diff
@@ -91,4 +92,6 @@ public struct APNSPushType: Hashable {
     ///
     /// - Important: If you set this push type, you must use the topic from the UID attribute in the subject of your MDM push certificate.
     public static let mdm = Self(configuration: .mdm)
+    
+    public static let liveactivity = Self(configuration: .liveactivity)
```
FranzBusch, line 96: Taking from the Apple docs
```suggestion
    /// Use the live activity push type to update your live activity.
    ///
    /// - Important: If you set this push type, the topic must use your app’s bundle ID with `.push-type.liveactivity` appended to the end.
    public static let liveactivity = Self(configuration: .liveactivity)
```

Sources/APNSwift/LiveActivity/APNSClient+LiveActivity.swift
```diff
@@ -0,0 +1,44 @@
...
+    ///
+    /// - Parameters:
+    ///   - notification: The notification to send.
+    ///
+    ///   - deviceToken: The hexadecimal bytes that identify the user’s device. Your app receives the bytes for this device token
+    ///    when registering for remote notifications.
```
FranzBusch, line 18: This is not true according to the Apple docs
```suggestion
    ///   - deviceToken: The hexadecimal bytes that identify the user’s device. Your apps receives the bytes after you started the live activity.
    ///    when registering for remote notifications.
```

Sources/APNSwift/LiveActivity/APNSClient+LiveActivity.swift
```diff
@@ -0,0 +1,44 @@
...
+        logger: Logger = _noOpLogger
+    ) async throws -> APNSResponse {
+        return try await self.send(
+            payload: notification,
+            deviceToken: deviceToken,
+            pushType: .liveactivity,
+            apnsID: notification.apnsID,
+            expiration: notification.expiration,
+            priority: notification.priority,
+            topic: notification.topic + ".push-type.liveactivity",
```
FranzBusch, line 38: I think this is the wrong place to append this. Could you take a look at how the `APNSComplicationNotification` or `APNSLocationNotification` are doing it.

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -0,0 +1,85 @@
...
+        self.priority = priority
+        self.topic = topic
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        // First we encode the user payload since this might use the `aps` key
+        // and we override it afterward.
+//        try self.payload.encode(to: encoder)
```
FranzBusch, line 82: This comment and code can be removed
```suggestion
```

Sources/APNSwift/LiveActivity/APNSLiveActivityNotificationAPSStorage.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  APNSLiveActivityNotificationAPSStorage.swift
+//  PushSender
+//
+//  Created by csms on 20/09/2022.
+//
+
+import Foundation
+
+public struct APNSLiveActivityNotificationAPSStorage<ContentState: Encodable>: Encodable {
```
FranzBusch, line 10: This must not be public and all its members

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -0,0 +1,85 @@
...
+        topic: String,
+        apnsID: UUID? = nil,
+        contentState: ContentState,
+        event: LiveActivityNotificationEvent,
+        timestamp: Int
+    ) {
+        self.aps = APNSLiveActivityNotificationAPSStorage(
+            timestamp: timestamp,
+            event: event.rawValue,
+            contentState: contentState
```
FranzBusch, line 69: We should expose these three properties as `public var` properties on this type as well to and add proper documentation.

Sources/APNSwift/LiveActivity/LiveActivityNotificationEvent.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  LiveActivityNotificationEvent.swift
+//  PushSender
+//
+//  Created by csms on 20/09/2022.
+//
+
+import Foundation
+
+public enum LiveActivityNotificationEvent: String {
```
FranzBusch, line 10: This type need storage be public since we have to expose a var for users to set but we can't publicly expose an `enum` since we can't add new cases without breaking API. Please follow the approach that `APNSExpiration` or `APNSPriority` takes and use a `public struct` with an `internal enum` inside. Furthermore could you add some docs here please.

FranzBusch (commented):
Thanks for adding this just left a few comments that I spotted when comparing it to the Apple docs and the previous changes to the other APIs

Sources/APNSwift/LiveActivity/LiveActivityNotificationEvent.swift
frogg: LiveActivityNotificationEvent should also contain `end`

nasircsms: Suggested changes are done.

Sources/APNSwift/LiveActivity/APNSClient+LiveActivity.swift
```diff
@@ -0,0 +1,44 @@
+//
```
kylebrowning, line 1: In these 4 new files, please mimic/copy the other files with their licensed headers.

You will receive proper attribution via the contributors list

Sources/APNSwift/LiveActivity/APNSClient+LiveActivity.swift
```diff
@@ -0,0 +1,44 @@
...
+    ///    from `pushTokenUpdates` async property of a live activity.
+    ///
+    ///   - deadline: Point in time by which sending the notification to APNs must complete.
+    ///
+    ///   - logger: The logger to use for sending this notification.
+    @discardableResult
+    @inlinable
+    public func sendLiveActivityNotification<ContentState: Encodable>(
+        _ notification: APNSLiveActivityNotification<ContentState>,
+        activityPushToken: String,
```
kylebrowning, line 27:
```suggestion
        deviceToken: String,
```

kylebrowning: This should align with our other API's

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -0,0 +1,139 @@
...
+//
+//  Created by csms on 20/09/2022.
+//
+
+import struct Foundation.UUID
+
+/// An alert notification.
+///
+/// - Important: Your dynamic payload will get encoded to the root of the JSON payload that is send to APNs.
+/// It is **important** that you do not encode anything with the key `aps`.
```
kylebrowning, line 13: These are not necessary for this file
```suggestion
```

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -0,0 +1,139 @@
...
+public struct APNSLiveActivityNotification<ContentState: Encodable>: Encodable {
+    enum CodingKeys: CodingKey {
+        case aps
+    }
+
+    /// The fixed content to indicate that this is a background notification.
+    private var aps: APNSLiveActivityNotificationAPSStorage<ContentState>
+
+    // Timestamp when sending notification
```
kylebrowning, line 22:
```suggestion
    /// Timestamp when sending notification
```

Is there anything else we can say about what this timestamp does?

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -0,0 +1,139 @@
...
+        get {
+            return self.aps.timestamp
+        }
+        
+        set {
+            self.aps.timestamp = newValue
+        }
+    }
+    
+    // Event type e.g. update
```
kylebrowning, line 33:
```suggestion
    /// Event type e.g. update
```

I believe this could be elaborated on as well.

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -0,0 +1,139 @@
...
+        get {
+            return LiveActivityNotificationEvent(rawValue: self.aps.event)
+        }
+        
+        set {
+            self.aps.event = newValue.rawValue
+        }
+    }
+    
+    // Updated content-state of live activity
```
kylebrowning, line 44:
```suggestion
    /// The dynamic content of a Live Activity.
```

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -0,0 +1,139 @@
...
+    ///   - timestamp: Timestamp when sending notification
+    ///   - event: event type e.g. update
+    public init(
+        expiration: APNSNotificationExpiration,
+        priority: APNSPriority,
+        appID: String,
+        apnsID: UUID? = nil,
+        contentState: ContentState,
+        event: LiveActivityNotificationEvent,
+        timestamp: Int
```
kylebrowning, line 96:
```suggestion
        expiration: APNSNotificationExpiration,
        priority: APNSPriority,
        appID: String,
        contentState: ContentState,
        event: LiveActivityNotificationEvent,
        timestamp: Int,
        apnsID: UUID? = nil,
```

kylebrowning: apnsID should be last since it is optional

Sources/APNSwift/LiveActivity/APNSLiveActivityNotificationAPSStorage.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  APNSLiveActivityNotificationAPSStorage.swift
+//  PushSender
+//
+//  Created by csms on 20/09/2022.
+//
+
+import Foundation
```
kylebrowning, line 8:
```suggestion
```

Sources/APNSwift/LiveActivity/LiveActivityNotificationEvent.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  LiveActivityNotificationEvent.swift
+//  PushSender
+//
+//  Created by csms on 20/09/2022.
+//
+
+import Foundation
```
kylebrowning, line 8:
```suggestion
```

Sources/APNSwift/LiveActivity/LiveActivityNotificationEvent.swift
```diff
@@ -0,0 +1,21 @@
...
+
+public struct LiveActivityNotificationEvent: Hashable {
```
kylebrowning, line 10:
```suggestion
public struct APNSLiveActivityNotificationEvent: Hashable {
```

Sources/APNSwift/LiveActivity/LiveActivityNotificationEvent.swift
```diff
@@ -0,0 +1,21 @@
...
+    
+    /// The underlying raw value that is send to APNs.
+    @usableFromInline
+    internal let rawValue: String
+    
+    // Specifies that live activity should be updated
```
kylebrowning, line 16:
```suggestion
    /// Specifies that live activity should be updated
```

Sources/APNSwift/LiveActivity/LiveActivityNotificationEvent.swift
```diff
@@ -0,0 +1,21 @@
...
+    public static let update = Self(rawValue: "update")
+    
+    // Specifies that live activity should be ended
```
kylebrowning, line 19:
```suggestion
    /// Specifies that live activity should be ended
```

kylebrowning (changes requested):
Thanks so much, some more fixes here.

Along with 👇🏻 
- The topic property should include that the suffix is required same as the location notification does
- There are mentions of the `APNSAlertNotification` inside the new `APNSLiveActivityNotificaiton`

kylebrowning:
@nasircsms planning n making these changes or should I take over?

kylebrowning:
closing in favor of #164

PR #166, opened 2022-11-30
bugfix: when APNSClient deinit, syncShutdown HTTPClient

wangzhizhou:
image

### DIFF
--- a/Sources/APNSwift/APNSClient.swift
+++ b/Sources/APNSwift/APNSClient.swift
@@ -134,6 +134,10 @@ public final class APNSClient<Decoder: APNSJSONDecoder, Encoder: APNSJSONEncoder
         self.backgroundActivityLogger.trace("APNSClient is shutting down")
         try self.httpClient.syncShutdown()
     }
+    
+    deinit {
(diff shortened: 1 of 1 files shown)


Sources/APNSwift/APNSClient.swift
```diff
@@ -134,6 +134,10 @@ public final class APNSClient<Decoder: APNSJSONDecoder, Encoder: APNSJSONEncoder
         self.backgroundActivityLogger.trace("APNSClient is shutting down")
         try self.httpClient.syncShutdown()
     }
+    
+    deinit {
```
FranzBusch, line 138: The problem you are seeing can be fixed by calling shutdown or syncShutdown in your code. Doing this in deinit ist not safe.

FranzBusch (changes requested):

FranzBusch:
Thanks for opening this. The problem you are seeing is due to the fact that your code is dropping the reference to the APNSClient without calling shutdown. It is not something we can fix here

kylebrowning:
beat me too it. Yes you should be managing shutting down, not us

PR #167, opened 2023-01-02
Add support for dismissal-date to live activity notification

mark-urbanthings:
The existing implementation didn't have support for `dismissal-date` so have added support to the current beta. Code is working in prototype Vapor live notification service we've been working on. Changes:

- Updated APNSLiveActivityNotification to accept a dismissal date as unix timestamp value.
- Added unit tests to validate JSON encoding of APNSLiveActivityNotification.
- Added extension to Int to provide `dismissImmediately` constant.
- Updated examples in `README.md`

Would be helpful if this accepted - thanks!


Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -78,7 +88,7 @@ public struct APNSLiveActivityNotification<ContentState: Encodable>: Encodable {
 
     /// The topic for the notification. In general, the topic is your app’s bundle ID/app ID.
     public var topic: String
-
+
```
kylebrowning, line 91:
```suggestion

```

kylebrowning (approved):
This LGTM, will wait for tests to pass

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -13,8 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 import struct Foundation.UUID
+import struct Foundation.Date
```
FranzBusch, line 16:
```suggestion
```

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -58,6 +59,15 @@ public struct APNSLiveActivityNotification<ContentState: Encodable>: Encodable {
         }
     }
  
+    public var dismissalDate: Int? {
```
FranzBusch, line 62: Instead of basing this on `Int` can we please introduce a new type akin to the `APNSNotificationExpiration`. This will make it clear what this type is for and avoids an extension on `Int`. I would recommend having a `static let immediately` and a `static func timeIntervalSince1970InSeconds(_ timeInterval: Int)`.

Sources/APNSwift/LiveActivity/APNSLiveActivityNotificationAPSStorage.swift
```diff
@@ -12,25 +12,30 @@
 //
 //===----------------------------------------------------------------------===//
 
+import struct Foundation.Date
```
FranzBusch, line 15:
```suggestion
```

FranzBusch (commented):
Thanks for adding this. Left some inline comments

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
mark-urbanthings: Makes sense, have committed changes for this

Sources/APNSwift/LiveActivity/APNSLiveActivityNotificationAPSStorage.swift
mark-urbanthings: Well spotted, had Date in use earlier but redundant now

Sources/APNSwift/LiveActivity/APNSLiveActivityNotification.swift
kylebrowning: @mark-urbanthings Looks like there is still this added space here

mark-urbanthings: Fixed

kylebrowning:
Thanks @mark-urbanthings

PR #170, opened 2023-02-14
Breakout APNSwift into packages

kylebrowning:
This is the first PR among probably many to break out APNS into APNSCore, APNSAHC, APNSUrlSession, and APNS(a hyper optimized h2 client)

PR #174, opened 2023-04-10
Make APNSError.reason public again

vojtarylko:
This commit restores public access to the `APNSError.reason` property, because it's needed for library users handling APNs errors.  (The status code is not an viable alternative, as there are multiple different error reasons for one status code.)

The `reason` property was unintentionally set to private in commit ecf0f4465 (Breakout APNSwift into packages (#170)). Notably, the associated type ErrorReason remained public.


0xTim (approved):
This seems fine to me, any objections @kylebrowning ?

kylebrowning:
yup! I was talking with Franz about this just to understand why we did it in the first place, but this pr can be merged!

Issue #179, opened 2023-08-02
HOW to send raw custom payload


SupaBasti:
the README says it would be able to send Data as raw payload, but I do not find the proper API. There is one send method with takes an APNSRequest as parameter with a message property conforming to APNSMessage protocol.

However, often the custom payload has different properties and thus cannot be represented by a struct with fix property types.

So, how can I add a custom Payload dictionary to an APNSMessage?

kylebrowning:
@SupaBasti Can you extend APNSMessage and send that?

```swift
public struct MyCustomMessage<Payload: Encodable & Sendable>: APNSMessage, Sendable {

kylebrowning:
I suppose not. let me take a look at the API

kylebrowning:
Can you give me an example of what you're trying to accomplish ?

SupaBasti:
I found a workaround: I just send my custom payload as [String:String] dictionary instead of [String: Any]. Thus I can put it directly into the payload property. All fine. Thanks

PR #185, opened 2023-10-11
support send raw payload data

wangzhizhou:
such as the json string payload could be transform to data and be sent 

If the rawPayloadData be set, message would not been encoded into bytebuffer, the rawPayloadData as the whole post request body bytes.

If rawPayloadData is nil, the send logic is same as the 5.0.0, that is the message been encoded into post request body and be sent.

There is an example use it:
```swift
      var client = APNSClient(
          configuration: .init(
              authenticationMethod: .jwt(
                  privateKey: try .init(pemRepresentation: config.privateKey),
                  keyIdentifier: config.keyIdentifier,
                  teamIdentifier: config.teamIdentifier
              ),
              environment: config.apnsServerEnv.env
          ),
          eventLoopGroupProvider: .createNew,
          responseDecoder: JSONDecoder(),
          requestEncoder: JSONEncoder()
      )

      guard let payloadData = """
      {
        "aps": {
          "alert": {
            "body": "Body Content",
            "subtitle": "Subtitle Content",
            "title": "Simple Alert Content"
          }
        },
        "o_url": "https://www.baidu.com",
        "obj": {
          "key1": "value",
          "key2": 2
        }
      }
      """.data(using: .utf8)
      else {
          return false
      }
      
      var response: APNSResponse?
      switch config.pushType {
      case .alert:
          response = try await client?.sendAlertNotification(
              .init(
                  alert: .init(
                      title: .raw("Simple Alert"),
                      subtitle: .raw("Subtitle"),
                      body: .raw("Body")),
                  expiration: .immediately,
                  priority: .immediately,
                  topic: config.appBundleID,
                  rawPayloadData: payloadData
              ),
              deviceToken: config.deviceToken)
      case .background:
          response = try await client?.sendBackgroundNotification(
              .init(
                  expiration: .immediately,
                  topic: config.appBundleID,
                  rawPayloadData: payloadData
              ),
              deviceToken: config.deviceToken)
      case .voip:
          response = try await client?.sendVoIPNotification(
              .init(
                  priority: .immediately,
                  appID: config.appBundleID,
                  rawPayloadData: payloadData
              ),
              deviceToken: config.pushKitVoIPToken)
      case .fileprovider:
          response = try await client?.sendFileProviderNotification(
              .init(
                  expiration: .immediately,
                  appID: config.appBundleID,
                  rawPayloadData: payloadData
              ),
              deviceToken: config.pushKitFileProviderToken)
      }

```

### DIFF
--- a/Sources/APNS/APNSClient.swift
+++ b/Sources/APNS/APNSClient.swift
@@ -142,71 +142,78 @@ public final class APNSClient<Decoder: APNSJSONDecoder, Encoder: APNSJSONEncoder
 extension APNSClient {
     
     public func send(_ request: APNSCore.APNSRequest<some APNSCore.APNSMessage>) async throws -> APNSCore.APNSResponse {
+        
+        // Device token
+        let deviceToken = request.deviceToken;
+        
+        // Headers
+        let headers = try await headers(of: request)
+        
+        // Payload ByteBuffer
+        var byteBuffer = self.byteBufferAllocator.buffer(capacity: 0)
+        
+        // Write Payload Data into Payload ByteBuffer
+        if let rawPayloadData = request.rawPayloadData {
+            byteBuffer.writeBytes(rawPayloadData)
+        } else {
+            try self.requestEncoder.encode(request.message, into: &byteBuffer)
+        }
+        
+        // Send Payload with deviceToken and Request Headers
+        let response = try await send(with: deviceToken, headers: headers, payload: byteBuffer)
+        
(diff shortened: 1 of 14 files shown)


kylebrowning:
This is not something we are going to support going forward, if you want to send a raw payload, you can do so as mentioned here. https://github.com/swift-server-community/APNSwift/issues/179

wangzhizhou:
Get it, Thx

PR #187, opened 2023-11-20
Update APNSClient+Location.swift

DominatorVbN:
Marked sendLocationNotification public to be accessible outside of APNSCore


0xTim:
@kylebrowning is there any reason this was internal?

kylebrowning:
Pretty sure it was just an oversight.

0xTim (approved):

PR #190, opened 2024-01-08
Fixed broken links in README.md

ladiesman218:
Appended `--end` to link destination for [Sending Live Activity Update]

Removed 2 links:
[Foundations](#foundations)
[Using the non semantic safe APIs](#using-the-non-semantic-safe-apis) Contents are missing in current readme.


0xTim (approved):
Thanks!

PR #191, opened 2024-01-25
adds start

2020Deception:


alex-taffe:
Plus one for this

BastianKusserow:
This is not enough to support live activity start notifications. The API would also have to support the activity attributes as well as the attributes type.

kylebrowning:
Hmm yeah I think we need just a little bit more here. Let me see what I can do.

PR #196, opened 2024-04-03
Add ability to start live activities

nakajima:
I believe this PR brings things in line with the JSON payload described in [the docs](https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications#Construct-the-payload-that-starts-a-Live-Activity).

This makes the `APNSLiveActivityNotificationEvent` type a bit more clunky to deal with, so I'd love to hear if anyone has any ideas on how to improve that.

I also ran `swift-format --configuration .swift-format` on my changes which seems to have brought in some unrelated format changes. If this is a PR worth pursuing I'm happy to clean those up.

See also https://github.com/swift-server-community/APNSwift/pull/191.

### DIFF
--- a/Sources/APNSCore/LiveActivity/APNSLiveActivityNotification.swift
+++ b/Sources/APNSCore/LiveActivity/APNSLiveActivityNotification.swift
@@ -38,6 +38,16 @@ public struct APNSLiveActivityNotification<ContentState: Encodable & Hashable & 
         }
     }
 
+    public var alert: APNSAlertNotificationContent? {
+        get {
+            return self.aps.alert
+        }
+
+        set {
+            self.aps.alert = newValue
+        }
+    }
+
     /// Event type e.g. update
     public var event: any APNSLiveActivityNotificationEvent {
         get {
@@ -48,14 +58,17 @@ public struct APNSLiveActivityNotification<ContentState: Encodable & Hashable & 
                 return APNSLiveActivityNotificationEventUpdate()
             default:
                 guard let attributesType = self.aps.attributesType,
-                    let state = self.aps.attributesContent
+                    let state = self.aps.attributesContent,
+                    let alert = self.aps.alert
                 else {
                     // Default to update
                     return APNSLiveActivityNotificationEventUpdate()
                 }
 
                 return APNSLiveActivityNotificationEventStart(
-                    attributes: .init(type: attributesType, state: state))
+                    attributes: .init(type: attributesType, state: state),
+                    alert: alert
+                )
             }
         }
 
@@ -173,8 +186,7 @@ public struct APNSLiveActivityNotification<ContentState: Encodable & Hashable & 
 
         self.aps = APNSLiveActivityNotificationAPSStorage(
             timestamp: timestamp,
-            event: event.rawValue,
-            attributes: attributes,
+            event: event,
             contentState: contentState,
             dismissalDate: dismissalDate.dismissal
         )
--- a/Sources/APNSCore/LiveActivity/APNSLiveActivityNotification.swift
+++ b/Sources/APNSCore/LiveActivity/APNSLiveActivityNotification.swift
@@ -13,11 +13,12 @@
 //===----------------------------------------------------------------------===//
 
 import struct Foundation.UUID
+import struct Foundation.Data
 
 /// A live activity notification.
 ///
 /// It is **important** that you do not encode anything with the key `aps`.
-public struct APNSLiveActivityNotification<ContentState: Encodable>: APNSMessage {
+public struct APNSLiveActivityNotification<ContentState: Encodable & Hashable & Sendable>: APNSMessage {
     enum CodingKeys: CodingKey {
         case aps
     }
@@ -37,16 +38,28 @@ public struct APNSLiveActivityNotification<ContentState: Encodable>: APNSMessage
     }
     
     /// Event type e.g. update
-    public var event: APNSLiveActivityNotificationEvent {
+    public var event: any APNSLiveActivityNotificationEvent {
         get {
-            return APNSLiveActivityNotificationEvent(rawValue: self.aps.event)
+					switch self.aps.event {
+					case "end":
+						return APNSLiveActivityNotificationEventEnd()
+					case "update":
+						return APNSLiveActivityNotificationEventUpdate()
+					default:
+						guard let attributesType = self.aps.attributesType, let state = self.aps.attributesContent else {
+							// Default to update
+							return APNSLiveActivityNotificationEventUpdate()
+						}
+
+						return APNSLiveActivityNotificationEventStart(attributes: .init(type: attributesType, state: state))
+					}
         }
         
         set {
             self.aps.event = newValue.rawValue
         }
     }
-    
+
     /// The dynamic content of a Live Activity.
     public var contentState: ContentState {
         get {
@@ -57,7 +70,7 @@ public struct APNSLiveActivityNotification<ContentState: Encodable>: APNSMessage
             self.aps.contentState = newValue
         }
     }
- 
+
     public var dismissalDate: APNSLiveActivityDismissalDate? {
         get {
             return .init(dismissal: self.aps.dismissalDate)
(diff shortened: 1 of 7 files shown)


nakajima:
Ok actually it looks like I need to include an `alert` as well. On it.

nakajima:
OK, I've verified that this is working in a test app so it's ready for review!

Sources/APNSCore/LiveActivity/APNSLiveActivityNotificationEvent.swift
```diff
@@ -12,15 +12,65 @@
...
-    /// The underlying raw value that is send to APNs.
-    @usableFromInline
-    internal let rawValue: String
-    
-    /// Specifies that live activity should be updated
-    public static let update = Self(rawValue: "update")
-    
-    /// Specifies that live activity should be ended
-    public static let end = Self(rawValue: "end")
+public protocol APNSLiveActivityNotificationEvent: Hashable, Encodable {
```
FranzBusch, line 15: This is a breaking change and we can't just change from a struct to a protocol. We should try and continue to model it with the current struct type.

Sources/APNSCore/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -17,7 +17,9 @@ import struct Foundation.UUID
 /// A live activity notification.
 ///
 /// It is **important** that you do not encode anything with the key `aps`.
-public struct APNSLiveActivityNotification<ContentState: Encodable>: APNSMessage {
+public struct APNSLiveActivityNotification<ContentState: Encodable & Hashable & Sendable>:
```
FranzBusch, line 20: We can't just add these requirements to the generic type. This is a breaking API change otherwise. We can make the conformance to those protocols conditional though and that should work.

Sources/APNSCore/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -30,34 +32,62 @@ public struct APNSLiveActivityNotification<ContentState: Encodable>: APNSMessage
...
-    public var event: APNSLiveActivityNotificationEvent {
+    public var event: any APNSLiveActivityNotificationEvent {
         get {
-            return APNSLiveActivityNotificationEvent(rawValue: self.aps.event)
+            switch self.aps.event {
+            case "end":
+                return APNSLiveActivityNotificationEventEnd()
+            case "update":
+                return APNSLiveActivityNotificationEventUpdate()
+            default:
```
FranzBusch, line 59: We should be exhaustive here and avoid using default cases if possible.

FranzBusch (commented):
Thanks for working on this. I left a few comments on the PR. Mostly this is a breaking API change so we have to do it slightly differently to avoid that breakage.

nakajima:
@FranzBusch Thanks so much for the review! I've moved things around to just introduce a separate object for the additional data a `start` event needs. I think the one downside to this approach is that if you pass a `start` event without the `startOptions`, the push just won't get sent. I don't think that's the end of the world though since that's also what APNS does afaik.

> We can't just add these requirements to the generic type. This is a breaking API change otherwise. We can make the conformance to those protocols conditional though and that should work.

Fixed in https://github.com/swift-server-community/APNSwift/pull/196/commits/5af6d92d57faddf3e9412303a8df81e383a87299 and https://github.com/swift-server-community/APNSwift/pull/196/commits/fffc9eaa7f9eaa95892c2e6c52080773e08f43d8.

> This is a breaking change and we can't just change from a struct to a protocol. We should try and continue to model it with the current struct type.

Fixed in https://github.com/swift-server-community/APNSwift/pull/196/commits/4671d0ada8a65122fd7ee33e8916d5ec5439bf15.

nakajima:
Actually this PR is wrong. I think we need to pass the ActivityAttributes as the `attributes` field, not the content state.

Sources/APNSCore/LiveActivity/APNSLiveActivityNotification.swift
```diff
@@ -152,7 +151,7 @@ public struct APNSLiveActivityNotification<ContentState: Encodable>: APNSMessage
     ) {
         self.aps = APNSLiveActivityNotificationAPSStorage(
             timestamp: timestamp,
-            event: event.rawValue,
+						event: event.rawValue,
```
alex-taffe, line 154: Just before this gets merged, funky indentation here

nakajima: Fixed in https://github.com/swift-server-community/APNSwift/pull/196/commits/2d41546e2d7ffd6df8c2bba7bd3c25cf8ce7c3c2.

Sources/APNSCore/LiveActivity/APNSStartLiveActivityNotification.swift
```diff
@@ -0,0 +1,176 @@
...
+//===----------------------------------------------------------------------===//
+
+import struct Foundation.UUID
+
+/// A notification that starts a live activity
+///
+/// It is **important** that you do not encode anything with the key `aps`.
+public struct APNSStartLiveActivityNotification<Attributes: Encodable, ContentState: Encodable>:
+    APNSMessage
+{
```
FranzBusch, line 22: This formatting looks weird. Can we fix this up?

Sources/APNSCore/LiveActivity/APNSStartLiveActivityNotificationAPSStorage.swift
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of APNSwift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+struct APNSStartLiveActivityNotificationAPSStorage<Attributes: Encodable, ContentState: Encodable>:
```
FranzBusch, line 15: Same here

Sources/APNSCore/LiveActivity/APNSStartLiveActivityNotification.swift
```diff
@@ -0,0 +1,176 @@
...
+    ///   - appID: Your app’s bundle ID/app ID. This will be suffixed with `.push-type.liveactivity`.
+    ///   - apnsID: A canonical UUID that identifies the notification.
+    ///   - contentState: Updated content-state of live activity
+    ///   - timestamp: Timestamp when sending notification
+    ///   - dismissalDate: Timestamp when to dismiss live notification when sent with `end`, if in the past
+    ///    dismiss immediately
+    ///   - attributes: The ActivityAttributes of the live activity to start
+    ///   - attributesString: The type name of the ActivityAttributes you want to send
+    ///   - alert: An alert that will be sent along with the notification
+    public init(
```
FranzBusch, line 109: Do we need two inits?

FranzBusch (commented):
Looks good. Just one question

Sources/APNSCore/LiveActivity/APNSStartLiveActivityNotification.swift
nakajima: No I don't think we do, removed in https://github.com/swift-server-community/APNSwift/pull/196/commits/a142abee3531d8297a28d69e3a7f492c75cc3d10.

chrisschlitt:
Can confirm this works, was able to start a live activity from a push notification with:
```
let notification = APNSStartLiveActivityNotification(
    expiration: .immediately,
    priority: .immediately,
    appID: "com.example.example",
    contentState: TestContentState(),
    timestamp: 1_672_680_658,
    attributes: TestAttributes(),
    attributesType: "TestAttributes",
    alert: .init(title: .raw("Test"), body: .raw("Test"))
)
do {
    let response = try await apnsClient.send(
        APNSRequest(
            message: notification,
            deviceToken: kit.token,
            pushType: .liveactivity,
            expiration: notification.expiration,
            priority: notification.priority,
            apnsID: notification.apnsID,
            topic: notification.topic,
            collapseID: nil
        )
    )
    // Success
} catch {
    // Failure
}
```

alex-taffe:
I know this is somewhat off topic, but while everyone is here and I can solicit advice, [Apple's docs](https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications#Start-new-Live-Activities-with-ActivityKit-push-notifications) state:
> When the system receives the ActivityKit push notification on a device, it starts a new Live Activity, wakes up your app, and grants it background run time to allow you to download assets that the Live Activity needs.

Does anyone know what method is invoked or what happens when the start notification goes out? I'm severely doubting it's just `application:didFinishLaunchingWithOptions:` but I could be wrong. I'm going to play around with it because I need it to download image assets, but just figured I'd just check to see if anyone has already figured it out

PR #200, opened 2024-05-21
Fix crash when using AnyEncodable payload for alert and background notifications

grighakobian:
In this update, I've addressed an issue where using [AnyEncodable](https://github.com/Flight-School/AnyCodable/blob/master/Sources/AnyCodable/AnyEncodable.swift) for user-provided notification payloads caused a crash during alert or background notifications. These changes resolve the crash and ensure stable handling of `AnyEncodable` payloads.


wangzhizhou:
This Merge fixed the problem described in: https://github.com/Flight-School/AnyCodable/issues/80

wangzhizhou:
may we wait for an release version

wangzhizhou:
@grighakobian Can you describe the reason caused this problem? 
Change the encode method order resolved this problem?

PR #202, opened 2024-07-17
Adjust to Swift 6 toolchain compile

wangzhizhou:
use the build command provided by Swift Package Index below to verify this package can be migrated to Swift 6 gracefully

```bash
env DEVELOPER_DIR=/Applications/Xcode-16.0.0-Beta.3.app xcrun swift build --arch arm64 -Xswiftc -strict-concurrency=complete
```

Reference this WWDC24 video to make change: https://developer.apple.com/wwdc24/10169

and the Guide to Migrate to Swift 6 on Swift Package Index: https://swiftpackageindex.com/ready-for-swift-6

If this PR is merged, after triggering the Swift Package Index build workflow, you can see the main branch be Safe from data race logo on the homepage of this package as below:

image

### DIFF
--- a/Sources/APNS/APNSClient.swift
+++ b/Sources/APNS/APNSClient.swift
@@ -125,7 +125,7 @@ public final class APNSClient<Decoder: APNSJSONDecoder, Encoder: APNSJSONEncoder
     /// - Parameters:
     ///   - queue: The queue on which the callback is invoked on.
     ///   - callback: The callback that is invoked when everything is shutdown.
-    public func shutdown(queue: DispatchQueue = .global(), callback: @escaping (Error?) -> Void) {
+    public func shutdown(queue: DispatchQueue = .global(), callback: @Sendable @escaping (Error?) -> Void) {
         self.httpClient.shutdown(callback)
     }
 
--- a/Sources/APNS/APNSClient.swift
+++ b/Sources/APNS/APNSClient.swift
@@ -125,7 +125,7 @@ public final class APNSClient<Decoder: APNSJSONDecoder, Encoder: APNSJSONEncoder
     /// - Parameters:
     ///   - queue: The queue on which the callback is invoked on.
     ///   - callback: The callback that is invoked when everything is shutdown.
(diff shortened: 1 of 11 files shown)


Sources/APNS/APNSClient.swift
```diff
@@ -125,7 +125,7 @@ public final class APNSClient<Decoder: APNSJSONDecoder, Encoder: APNSJSONEncoder
     /// - Parameters:
     ///   - queue: The queue on which the callback is invoked on.
     ///   - callback: The callback that is invoked when everything is shutdown.
-    public func shutdown(queue: DispatchQueue = .global(), callback: @escaping (Error?) -> Void) {
+    public func shutdown(queue: DispatchQueue = .global(), callback: @Sendable @escaping (Error?) -> Void) {
```
0xTim, line 128: Unless we're going for a major version, this function should be annotated `@preconcurrency`

0xTim (changes requested):
Thanks! One change requested

Sources/APNS/APNSClient.swift
wangzhizhou: OK, I will make this change as you described

wangzhizhou: @0xTim After I have a try, this @preconcurrency attribute can not be added in this scene:

image

or 

image

wangzhizhou: Are there any other suggestions for implementing this change?

0xTim: Sorry `@preconcurrency` goes before `func` - e.g. https://github.com/vapor/vapor/blob/main/Sources/Vapor/Utilities/Thread.swift#L5

wangzhizhou: Ok，I have never used it before， may I have a try later.thanks Tim

wangzhizhou: @0xTim I have add the @preconcurrency before the func shutdown

PR #181, opened 2023-09-02
Push to Talk (PTT) Support

mash3l777:
👋 Hello World,

Summary:

This pull request introduces support for the new Push to Talk (PTT) framework available in iOS 16.0+ and iPadOS 16.0+. Check [Apple Documentation](https://developer.apple.com/documentation/pushtotalk) for more info. 

Changes:

Added a APNSPushToTalkNotification Structure.
Updated a APNSClientProtocol with sendPushToTalkNotification function.
Added the new APNSPushType `.pushtotalk` 🎉.
Updated APNSExample -> Program.swift to reflect these changes.

Why this is important:

The Push to Talk framework is a game-changer for real-time audio transmission. Its power efficiency and user-friendly design . Integrating this into APNSSwift will make our library even more awesome for developers.

Testing:

I've added unit tests to cover this new functionality and also performed manual tests to ensure everything is working as expected.

Looking forward to your feedback! 😊

Cheers!☕️

PR #206, opened 2024-08-20
Privatize APNSPushType

kylebrowning:
This puts the `APNSPushType` back behind a private enum so that it won't be a breaking change when new ones are added.

### DIFF
--- a/Sources/APNS/APNSClient.swift
+++ b/Sources/APNS/APNSClient.swift
@@ -145,7 +145,7 @@ extension APNSClient {
         var headers = self.defaultRequestHeaders
 
         // Push type
-        headers.add(name: "apns-push-type", value: request.pushType.configuration.rawValue)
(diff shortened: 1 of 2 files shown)


0xTim (commented):
Are there changes for the configuration planned? Because this seems like a small change for a major release

kylebrowning:
@0xTim yes see https://github.com/swift-server-community/APNSwift/pull/181

kylebrowning:
specifically https://github.com/swift-server-community/APNSwift/pull/181#discussion_r1314111956

kylebrowning:
@0xTim fwiw I don't think we need to go do a major just yet, people can always pull master if they want right?

0xTim:
Oh yeah for sure, if you're planning a major you don't have to release until you want to. (Would be a good opportunity to remove all the NIO stuff and go full structured concurrency)

0xTim (approved):

PR #209, opened 2024-09-14
Add content-available option to alert

hyouuu:
Currently you can either send an alert or send a background notification, but often times apps want to be waken up while showing an alert so they can do background data fetch etc, so why not adding the option (surprised that nobody raised this question before)


kylebrowning:
YEs, reading that content-available only works for background pushes

Issue #222, opened 2025-07-01
Can we make APNSClient's Sendable conformance public?


mergesort:
**Is your feature request related to a problem? Please describe.**
I'm attempting to integrate APNSwift into my Vapor project, and my app follows a pattern where I store dependencies in a `StorageKey`. Due to this conformance, the type being added to the StorageKey must be marked `Sendable`.

```swift
public protocol StorageKey {
    /// The type of the stored value associated with this key type.
    associatedtype Value: Sendable
}
```

**Describe the solution you'd like**
I noticed this extension already exists, and both `JSONEncoder` and `JSONDecoder` are marked as `@unchecked Sendable` in Apple's frameworks, so assuming this is possible to do in APNSwift it should be a quick improvement.
```swift
extension APNSClient: Sendable where Decoder: Sendable, Encoder: Sendable {}
```

**Describe alternatives you've considered**
I started by making my own `@unchecked Sendable` wrapper before noticing that conformance already existed. I don't really find much value from it and it's kind of an ugly solution, but I could always fall back to it if needed.
```swift
final class SendableAPNSClient: @unchecked Sendable {
	let client: APNSClient<JSONDecoder, JSONEncoder>

	private init(client: APNSClient<JSONDecoder, JSONEncoder>) {
		self.client = client
	}
}
```

**Additional context**
Not much really, except to say that so far using the project has been a delight so thank you very much for that!

mergesort:
As noted in [this comment](https://github.com/swift-server-community/APNSwift/pull/223#issuecomment-3030100627), this issue isn't actually needed. 🤦🏻

PR #223, opened 2025-07-02
Making APNSClient's Sendable conformance public

mergesort:
Just a one-liner, but this was useful for me to inject APNSClient into Vapor's StorageKey as described in #222.


mergesort:
Sorry for this PR and issue, it looks like this was an Xcode issue all along. I implemented this change on my fork, and it seemed to work, but that turned out to be something related to build artifacts. After trying a clean build it's no longer working, and this error is definitely a bug on my part and should be closed. Worst of all, it looks like I should have never even opened this PR because things already work as I needed them to — not sure what was wrong before. 🤦🏻

PR #225, opened 2025-07-11
Adding Package.swift for Swift 5 versions

mergesort:
I needed to deploy this library to my Swift 5 server, so I had to make a new Package.swift that wasn't tied to Swift 6 to make it work.

The two changes are:
- `// swift-tools-version:6.0` -> `// swift-tools-version:5.10`
- Removed `swiftLanguageModes: [.v6]`

Let me know if any changes are needed, happy to accept all feedback if you have another preferred way of doing this.


kylebrowning:
Nope!

PR #226, opened 2025-08-04
Add `sound` property to APNSAlertNotificationContent

Myrronth:
If you want to customize the sound that is played when a Live Activity is alerting, the `sound` property has to be part of `APNSAlertNotificationContent` instead of `APNSLiveActivityNotification` (see the example payload in the [Starting and updating Live Activities with ActivityKit push notifications](https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications#Construct-the-ActivityKit-remote-push-notification-payload))

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -5,8 +5,9 @@ Please refer to the [Contributing](CONTRIBUTING.md) section for more information
 
 | Version | Supported          |
 | ------- | ------------------ |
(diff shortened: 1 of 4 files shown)


Myrronth:
@kylebrowning Missed that some tests were not passing, but it's resolved now.

PR #228, opened 2025-08-04
Remove dismissalDate from APNSStartLiveActivityNotification

Myrronth:
A `dismissalDate` can only be set when an update is sent with the `end` event. When setting it while sending with other events (start or update) it has no impact.


kylebrowning:
@Myrronth some conflicts here.

